PR #7, opened 2020-09-22
Bugfix/new currency fails at harvesine distance

clangenb:
* Fix overflow error  in `encointer_community::harvesine_distance` by changing the type of `Degree` to `I64F64`

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -44,12 +44,11 @@ use rstd::prelude::*;
 use codec::{Decode, Encode};
 use sp_runtime::traits::{CheckedSub, IdentifyAccount, Member, Verify};
 
-use encointer_primitives::communities::LossyFrom;
 use encointer_primitives::{
     balances::BalanceType,
     ceremonies::consts::{AMOUNT_NEWBIE_TICKETS, REPUTATION_LIFETIME},
     ceremonies::*,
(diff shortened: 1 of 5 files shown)


clangenb:
Issue in substrate fixed: https://github.com/encointer/substrate-fixed/issues/6

ceremonies/src/lib.rs
```diff
@@ -627,8 +628,8 @@ impl<T: Trait> Module<T> {
         let perdegree = day / T::Moment::from(360u32);
         let start = next - duration;
         // rounding to the lower integer degree. Max error: 240s = 4min
-        let abs_lon: i32 = mlocation.lon.abs().lossy_into();
-        let abs_lon_time = T::Moment::from(abs_lon as u32) * perdegree;
+        let abs_lon: u32 = i64::lossy_from(mlocation.lon.abs()).saturated_into();
```
brenzi, line 631: lon is signed! If unit tests pass, then we need better test coverage!

brenzi (changes requested):

ceremonies/src/lib.rs
clangenb: But we are taking the absolute value here?

brenzi: true that

PR #10, opened 2020-11-24
Bazaar pallet

BignaH:
Only store removal and upload implemented for now. Articles will be implemented after successful review.


bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
+    //type Public: IdentifyAccount<AccountId = Self::AccountId>;
+    //type Signature: Verify<Signer = Self::Public> + Member + Decode + Encode;
+}
+// Logger target
+const LOG: &str = "encointer";
+
+pub type ShopIdentifier = u64; //URL
+pub type ArticleIdentifier = u64; //URL
+/*
```
brenzi, line 68: please remove dead/commented code

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+*/
+
+
+decl_storage! {
+    trait Store for Module<T: Trait> as Bazaar {
+        // Maps the shop or article owner to the respective items 
+        pub ShopsOwned get(fn shops_owned): double_map hasher(blake2_128_concat) CurrencyIdentifier, hasher(blake2_128_concat) T::AccountId => Vec<ShopIdentifier>;
+        pub ArticlesOwned get(fn articles_owned): double_map hasher(blake2_128_concat) CurrencyIdentifier, hasher(blake2_128_concat) T::AccountId => Vec<ArticleIdentifier>; 
+        // Show item affiliation
+        pub ShopAffiliation get(fn shop_affiliation): double_map hasher(blake2_128_concat) CurrencyIdentifier, hasher(blake2_128_concat) ShopIdentifier => T::AccountId;
```
brenzi, line 86: Why not `ShopOwner` and ` `ArticleOwner` would match wording of `shopsOwned`

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        pub ArticleAffiliation get(fn article_affiliation): double_map hasher(blake2_128_concat) CurrencyIdentifier, hasher(blake2_128_concat) ArticleIdentifier => (T::AccountId, ShopIdentifier);
+        // The set of all shops and articles per currency (community)
+        pub ShopRegistry get(fn shop_registry): map hasher(blake2_128_concat) CurrencyIdentifier => Vec<ShopIdentifier>;
```
brenzi, line 89: This could become a very big Vec, but fir now that's fine I guess

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        pub ArticleRegistry get(fn article_registry): map hasher(blake2_128_concat) CurrencyIdentifier => Vec<ArticleIdentifier>;
+    }
+}
+
+decl_event! {
+    pub enum Event<T> where AccountId = <T as frame_system::Trait>::AccountId {
+        /// Event emitted when a shop is uploaded. [who, shop]
+        ShopCreated(AccountId, ShopIdentifier),
```
brenzi, line 97: I'd add the cid for easy filtering

bazaar/src/lib.rs
brenzi: the bad thing about this is that the entire vec needs to be loaded and stored with each mutation. The alternative would be a map-set:
https://substrate.dev/recipes/map-set.html
but, again, lets leave it as it is for now

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+            let mut owned_shops = ShopsOwned::<T>::get(cid, &sender);
+            let mut shops = ShopRegistry::get(cid);
+
+            // Verify that the specified shop has not already been created with fast search
+            ensure!(!ShopAffiliation::<T>::contains_key(cid, &shop), Error::<T>::ShopAlreadyCreated);   
+            
+             // TODO: Really necessary to do the binary search twice just to get index?
+            // Get the index of the last entry of the Shop vector          
+            match shops.binary_search(&shop) {
```
brenzi, line 136: why do you check membersip again? if `ShopAffiliation` doesn't have an entry for `shop`, neither will `shops`

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+                // If the search succeeds, the shop was already created
+                Ok(_) => Err(<Error<T>>::ShopAlreadyCreated.into()),
+                // If the search fails, the shop can be inserted into the owned list
+                Err(shop_registry_index) => {
+                    match owned_shops.binary_search(&shop) {
+                        Ok(_) => Err(<Error<T>>::ShopAlreadyCreated.into()), // should not be possible
+                        Err(onwed_shops_index) => {
```
brenzi, line 143: dito

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        #[weight = 10_000]
+        pub fn remove_shop(origin, cid:CurrencyIdentifier, shop: ShopIdentifier) -> DispatchResult {
+            // Check that the extrinsic was signed and get the signer.
+            let sender = ensure_signed(origin)?;
+
+            let mut owned_shops = ShopsOwned::<T>::get(cid, &sender);
+            let mut shops = ShopRegistry::get(cid);
+
+            // Verify that the specified shop is existing.
+            ensure!(ShopAffiliation::<T>::contains_key(cid, &shop), Error::<T>::NoSuchShop);
```
brenzi, line 170: why should the attempt to delete an inexisting shop cause an error? The result is the same

bazaar/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        // initialisation
+      //  type Error = Error<T>;
+        fn deposit_event() = default;
+
+        /// Allow a user to create a shop
+        #[weight = 10_000]
+        pub fn upload_shop(origin, cid: CurrencyIdentifier, shop: ShopIdentifier) -> DispatchResult {
+            // Check that the extrinsic was signed and get the signer
+            let sender = ensure_signed(origin)?;
+
```
brenzi, line 127: you should verify that the cid supplied is a registered currency in `encointer_currencies` at this point

bazaar/src/tests.rs
```diff
@@ -0,0 +1,352 @@
...
+        let shops = EncointerBazaar::shop_registry(cid);
+        let alices_shops = EncointerBazaar::shops_owned(cid, alice);
+        // assert that shop was added
+        assert!(shops.contains(&alice_shop));
+        assert!(alices_shops.contains(&alice_shop));
+        // assert that the shop is owned by alice
+        assert_eq!(EncointerBazaar::shop_affiliation(&cid, &alice_shop), alice);
+    });
+}
+
```
brenzi, line 210: TODO: create_new_shop_with_bad_cid_fails

.gitignore
```diff
@@ -1,10 +1,13 @@
 # Generated by Cargo
 # will have compiled files and executables
-/target/
+**/target/**
```
brenzi, line 3: this should not be needed if bazaar is member of the workspace in the root Cargo.toml. But it's fine to leave it

bazaar/Cargo.toml
```diff
@@ -0,0 +1,116 @@
...
+default-features = false
+package = "sp-runtime"
+version = "2.0.0"
+
+[dependencies.sp-core]
+default-features = false
+package = "sp-core"
+version = "2.0.0"
+
+[dependencies.balances]
```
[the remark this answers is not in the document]
brenzi: please do the same wit hother crates you don't need for --release

brenzi (changes requested):
looking good. just a few wishes..

PR #13, opened 2020-12-16
Bazaar pallet change items to vec<u8>

haerdib:
Change type of bazaar items from uint to vec<u8> to support string types

### DIFF
--- a/bazaar/src/lib.rs
+++ b/bazaar/src/lib.rs
@@ -48,8 +48,8 @@ pub trait Trait: frame_system::Trait
     type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
 }
 
-pub type ShopIdentifier = u64; 
-pub type ArticleIdentifier = u64; 
+pub type ShopIdentifier = Vec<u8>; 
+pub type ArticleIdentifier = Vec<u8>; 
 
 decl_storage! {
     trait Store for Module<T: Trait> as Bazaar {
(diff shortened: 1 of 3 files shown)


brenzi (changes requested):
your code looks good but you need to replace substrate dependencies in Cargo.toml. 
```
git = "https://github.com/paritytech/substrate.git" 
rev = "a208da16"
```
for all substrate crates (`sp-* frame-* pallet-*`) for compatibility with node 0.3.7

bazaar/src/lib.rs
```diff
@@ -94,7 +94,7 @@ decl_module! {
 
         /// Allow a user to create a shop
         #[weight = 10_000]
-        pub fn upload_shop(origin, cid: CurrencyIdentifier, shop: ShopIdentifier) -> DispatchResult {
+        pub fn new_shop(origin, cid: CurrencyIdentifier, shop: ShopIdentifier) -> DispatchResult {
```
brenzi, line 97: one more thing, sorry:
I could upload a 1GB `Vec<u8>` here. You have to constrain the size of this vec to the size needed for IPFS URI
this applies to all dispatchables with dynamic size arguments!

Please also add a test where the dispatch maust fail with an oversize argument

The alternative would be to make the weight dependen t on the size of arguments, but that is of no use to you because Vec will always have the same length if used properly

brenzi (changes requested):

bazaar/src/lib.rs
haerdib: I have included a max length and the respective test.

Though not yet relevant, but it might be in the eventually:
IPFS might use different hashing algorithms in the future, which will result in longer byte lenghts. Maybe we should include tests on the app side (i.e. prepended multihash length coincides with hash digest) and change to weight depended blockchain inclusion?

PR #18, opened 2021-01-14
Bootstrappers endorse newcomers

### LINKED ISSUE #4: meetup assignment should never assign more meetups than there are locations for a currency
reproduce:
1. register currency with one location and 12 bootstrappers
2. register everybody
3. 2 meetups will be assigned

it should moreover never happen that the bootstrapping ceremony has more than 1 meetup

### LINKED ISSUE #11: overbooked meetups assigned
growing a bot community reveals that sometimes, we overbook meetups. the rule is that every meetup is max 12.

```
Phase is now: ATTESTING                                                                                                                                                                                                              [5/1928]
cid is 3KwrP77WiSJ5doBs3Rvqgb8MDB4AYMacKvijrb2nVjaC
phase is ATTESTING
number of known accounts: 151
****** Performing 13 meetups
Performing meetup with 13 participants
Performing meetup with 13 participants
Performing meetup with 13 participants
Performing meetup with 13 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
Performing meetup with 11 participants
```

### LINKED ISSUE #16: allow bootstrappers to endorse newcomers for faster population growth
For security, only X% newbies are assigned to meetups (currently 25%). When starting a community, this seems overly restrictive and may frustrate people who'd like to join.
As long as the group of participants can be known to the bootstrappers directly, it should be acceptable to assign more newbies.

Goal: Allow to grow population quickly at the beginning.

Assuming every bootstrapper knows 50 people well enough ( without overlap with other bootstrappers ) we'll grant every bootstrapper 50 endorsing tickets.

```
B: registered bootstrappers
R: registered reputables, non-bootstrappers
E: registered endorsees
N: registered newbies
M: meetups to be assigned

# change newbie limit to 1/3 => newbie capacity 50% of reputables+bootstrappers 
N = random crop N to length min(|N|, floor(|R|/2)
n = |B| + |R| + |E| + |N|
|M| = ceil(n/12)

P = B + R + E + N

fill meetup slots assigning one at the time 
```

Maximum growth:
| ceremony | P |
|--------------|-----|
|1|12|
|2|12+12*50+12/2 = 618|
|3| 618 + 618/2 = 927 |
|..||
|9 (1 year)| 10'056|

this way we save ~10 ceremonies time, more than a year

clangenb:
Bootstrappers can now endorse newbies to facilitate faster community growth. Closes #16.
Closes #4: Fix assigning to many meetups.
Closes #11: Fix overbooked meetups


ceremonies/src/lib.rs
```diff
@@ -310,22 +321,42 @@ decl_module! {
...
-    where
-        AccountId = <T as frame_system::Trait>::AccountId,
-    {
-        ParticipantRegistered(AccountId),
+        #[weight = 10_000]
+        pub fn endorse_newcomer(origin, cid: CurrencyIdentifier, newby: T::AccountId) -> DispatchResult {
+            debug::RuntimeLogger::init();
+            let sender = ensure_signed(origin)?;
+
+            // this is done to simplify the logic as a trade-off for UX, subject to change?
```
clangenb, line 335: @brenzi shall we change that such that endorsements in the assigning/attesting phase will be effective for the ceremony with `currency_ceremony_index + 1`?

brenzi: that makes sense. it should be possible to endorse anytime, irrespective of the phase.

ceremonies/src/lib.rs
```diff
@@ -125,14 +125,16 @@ decl_storage! {
         ParticipantIndex get(fn participant_index): double_map hasher(blake2_128_concat) CurrencyCeremony, hasher(blake2_128_concat) T::AccountId => ParticipantIndexType;
         ParticipantCount get(fn participant_count): map hasher(blake2_128_concat) CurrencyCeremony => ParticipantIndexType;
         ParticipantReputation get(fn participant_reputation): double_map hasher(blake2_128_concat) CurrencyCeremony, hasher(blake2_128_concat) T::AccountId => Reputation;
+        // newbies granted a ticket from a bootstrapper for the next ceremony.
+        Endorsees get(fn endorsees): map hasher(blake2_128_concat) CurrencyCeremony => Vec<T::AccountId>;
```
brenzi, line 129: have you considered solving this with a new variant in the Reputation enum? Vec is ok as the size is limited, but actually, an endorsement is a form of reputation, no?

ceremonies/src/lib.rs
brenzi: What do we do if a newcomer misses the ceremony? should the endorsement be void then? I think we could live with that for now. the bootsrapper could re-endorse

ceremonies/src/lib.rs
```diff
@@ -310,11 +312,41 @@ decl_module! {
...
+            ensure!(<encointer_scheduler::Module<T>>::current_phase() == CeremonyPhaseType::REGISTERING,
+                "endorsing participants can only be done during REGISTERING phase");
+
+            ensure!(<encointer_currencies::Module<T>>::currency_identifiers().contains(&cid),
+                "CurrencyIdentifier not found");
+
+            ensure!(<encointer_currencies::Module<T>>::bootstrappers(&cid).contains(&sender),
+            "only bootstrappers can endorse newbies");
+
+            ensure!(<encointer_currencies::Module<T>>::bootstrapper_newby_tickets(&cid, &sender) > 0,
```
brenzi, line 336: newbie? is that wrong or both ok?

clangenb: Oops, you are right newby is incorrect, shall change that.

ceremonies/src/lib.rs
clangenb: Ah nice idea, I must admit I did not think of that. I can't think of a reason to not use an enum variant.

ceremonies/src/lib.rs
clangenb: This is what I thought. Someone who cannot attend after being endorsed is actually not very reputable?

But we could migrate them to (cid, cindex + 1) but then there is a chance that some endorsees might never turn up and we migrate them forever. Even if we solve it differently, we should have some kind of mechanism to track endorsees who never turn up to free unused storage, shouldn't we?

ceremonies/src/lib.rs
clangenb: Then we should think about how we want to handle now-shows of endorsees in the future. If we don't want to forfeit the endorsement, how do we know the ceremony_index at which he has been endorsed?

I guess, the registering participant would need to supply a 'Proof_of_Endorsement'?

brenzi: that ProofOfEndorsement would be the consequence of using an enum variant, yes. Your `vec` approach is more straightforward I must admit.

Let's consider the act of endorsing:

Your `vec` version means the endorsee shows a QR code to the endorser who then submits a tx. The endorsee registers like everyone else.

ProofOfEndorsement is slightly more involved: The endorsee has to know he's been endorsed and register differently.

I think this just adds complexity and doesn't help with sclability, as it is only used in the initial phase. Let's stick with your `vec` version

clangenb: Yes, I would opt for the simpler version too.

clangenb:
Ready to merge from my perspective

brenzi:
You seem to use an older nightly. I get:
```
error[E0282]: type annotations needed
   --> balances/src/lib.rs:147:5
    |
146 |         let elapsed_time_u32: u32 = elapsed_time_block_number.try_into().ok()
    |                                     ------------------------------------ this method call resolves to `std::result::Result<T, <Self as TryInto<T>>::Error>`
147 |             .expect("blockchain will not exceed 2^32 blocks; qed").try_into().unwrap();
    |              ^^^^^^ cannot infer type for type parameter `T`
    |
    = note: type must be known at this point

error: aborting due to previous error

For more information about this error, try `rustc --explain E0282`.
error: could not compile `pallet-encointer-balances`
```
May I ask you to fix that ugly type conversion there? My commit nails down the rust-toolchain, so you should get the same error

clangenb:
> You seem to use an older nightly. I get:
> 
> ```
> error[E0282]: type annotations needed
>    --> balances/src/lib.rs:147:5
>     |
> 146 |         let elapsed_time_u32: u32 = elapsed_time_block_number.try_into().ok()
>     |                                     ------------------------------------ this method call resolves to `std::result::Result<T, <Self as TryInto<T>>::Error>`
> 147 |             .expect("blockchain will not exceed 2^32 blocks; qed").try_into().unwrap();
>     |              ^^^^^^ cannot infer type for type parameter `T`
>     |
>     = note: type must be known at this point
> 
> error: aborting due to previous error
> 
> For more information about this error, try `rustc --explain E0282`.
> error: could not compile `pallet-encointer-balances`
> ```
> 
> May I ask you to fix that ugly type conversion there? My commit nails down the rust-toolchain, so you should get the same error

Done, but I was using a newer rust-toolchain. Hence, this might get back to us.

brenzi:
Your code is not compliant with and may still suffer from #11. You treat reputables the same as endorsees and then the indexing restarts to assign newbies. #16 actually states that you should assemble a Vec with all participants first, ordered by type. I must insist on that logic because yours could lead to suboptimal distribution of bootstrappers

ceremonies/src/lib.rs
```diff
@@ -384,55 +412,66 @@ impl<T: Trait> Module<T> {
...
+                } else if <Endorsees<T>>::contains_key((cid, cindex), &participant) {
+                    endorsees.push(participant)
                 } else {
                     newbies.push(participant);
                 }
             }
-            let mut n = reputables.len();
-            n += min(newbies.len(), n / 4);
+            let mut n = bootstrappers.len() + reputables.len();
+            n += min(newbies.len(), n / 3);
```
brenzi, line 441: + endorsees.len()

ceremonies/src/lib.rs
```diff
@@ -384,55 +412,66 @@ impl<T: Trait> Module<T> {
...
+
+            let mut all_participants = bootstrappers
+                .into_iter()
+                .chain(reputables.into_iter())
+                .chain(endorsees.into_iter())
+                .chain(newbies.into_iter());
+
             let n_meetups = n / 12 + 1;
```
brenzi, line 449: this is not the same as `ceil(n/12)` ! I'd rather have `ceil()`. If we have 12 participants, we want one meetup, not two

ceremonies/src/lib.rs
```diff
@@ -384,55 +412,66 @@ impl<T: Trait> Module<T> {
...
-                } else {
-                    debug::debug!(target: LOG, 
-                        "had to skip a newbie: {:?}", 
-                        p);
-                }
+
+            for p in all_participants {
+                debug::debug!(target: LOG, "had to skip a newbie: {:?}", p);
             }
             // purge meetups that are too small
```
brenzi, line 463: Is this still necessary with the above logic? can there be too-small meetups? We should rather cover this with unit tests I think (test all possible numbers of all types of participants from 0 to 12 each)?

ceremonies/src/lib.rs
brenzi: As we're already editing this line, could you fix #4 at the same time?

brenzi (changes requested):

ceremonies/src/lib.rs
clangenb: pushed a commit in this exact moment

ceremonies/src/lib.rs
clangenb: Yes, I agree we should remove that.
And yes, I believe we should test these combinations in unit test.

ceremonies/src/tests.rs
```diff
@@ -1540,30 +1900,32 @@ fn grow_population_works() {
...
         }
         run_to_next_phase();
         // ASSIGNING
         assert_eq!(EncointerCeremonies::meetup_count((cid, cindex)), 2);
         let meetup4_1 = EncointerCeremonies::meetup_registry((cid, cindex), 1);
         let meetup4_2 = EncointerCeremonies::meetup_registry((cid, cindex), 2);
 
-        // whitepaper III-B Rule 3: no more than 1/4 participants without reputation
-        assert!(meetup4_1.len() + meetup4_2.len() <= 13);
+        // whitepaper III-B Rule 3: no more than 1/3 participants without reputation
```
clangenb, line 1923: @brenzi what do you think?

currencies/src/lib.rs
```diff
@@ -93,10 +93,13 @@ const RADIANS_PER_DEGREE: U0F64 = U0F64::from_bits(0x0477D1A894A74E40);
 // in meters
 const MEAN_EARTH_RADIUS: I32F0 = I32F0::from_bits(0x006136B8);
 
+const INITIAL_NEWBIE_TICKETS: u8 = 50;
+
 decl_storage! {
     trait Store for Module<T: Trait> as EncointerCurrencies {
         Locations get(fn locations): map hasher(blake2_128_concat) CurrencyIdentifier => Vec<Location>;
         Bootstrappers get(fn bootstrappers): map hasher(blake2_128_concat) CurrencyIdentifier => Vec<T::AccountId>;
+        BootstrapperNewbieTickets get(fn bootstrapper_newbie_tickets): double_map hasher(blake2_128_concat) CurrencyIdentifier, hasher(blake2_128_concat) T::AccountId => u8;
```
brenzi, line 102: this should be with the ceremonies pallet, not currencies. reasoning: Cantillon has `currencies` onchain and `ceremonies` off-chain in STF

currencies/src/lib.rs
```diff
@@ -153,6 +156,11 @@ decl_module! {
             <CurrencyIdentifiers>::mutate(|v| v.push(cid));
             <Locations>::insert(&cid, &loc);
             <Bootstrappers<T>>::insert(&cid, &bootstrappers);
+
```
brenzi, line 159: this should be kept here, because ultimatlely, the number of newbie tickets may be a custom currency parameter

brenzi: hmm, actuelly, it doesn't matter. may be more intuitive to have this in ceremonies as well....

ceremonies/src/tests.rs
brenzi: I agree:

ceremony1:
6B + 14N => 9 assigned, 1 meetup
ceremony2:
6B+ 3R + 11N => 13 assigned, 2 meetups
ceremony3:
6B+ 7R + 7N => 19 assigned, 2 meetups

ceremonies/src/tests.rs
```diff
@@ -1540,30 +1900,32 @@ fn grow_population_works() {
...
+        // Fixme: I do not agree with these numbers. According to my calculation we should have:
+        // |N| = (6B + 7R) + min(7N, floor(13/2)) = 19 Participants
+        // Is proof generation broken?
```
brenzi, line 1926: as this test passes, there seems to be an error somewhere

currencies/src/lib.rs
clangenb: Sorry, I am confused. What do you refer to with `this`? If it is the bootstrappers, I would also vote to put them in the `Ceremonies` pallet. What does the `Currencies` pallet need them for?

Furthermore, we never iterate through them, only perform lookups.  This implies the `BootstrapperNewbieTickets` double_map can actually replace the `Bootrappers` map, thereby gaining a slight performance improvement, see [Maps as sets](https://substrate.dev/recipes/map-set.html).

ceremonies/src/lib.rs
brenzi: lets keep it as a single chance. the endorser can endorse again, but that'll cost another credit

ceremonies/src/lib.rs
```diff
@@ -447,7 +447,11 @@ impl<T: Trait> Module<T> {
                 .chain(endorsees.into_iter())
                 .chain(newbies.into_iter());
 
-            let n_meetups = n / 12 + 1;
+            let mut n_meetups = n / 12;
+            if n.rem_euclid(12) > 0 {
```
brenzi, line 457: Please read #4 carefully! This commit does NOT fix #4. You need to fetch the number of locations defined for this community and limit the number of meetups to that number!

clangenb: Oh, sorry for that. I did indeed not read it carefully enough.

ceremonies/src/tests.rs
```diff
@@ -1778,26 +1779,27 @@ fn grow_population_works() {
     });
 }
 
-#[test]
-fn assign_meetup_works() {
+#[rstest(n_bootstrappers, n_reputables, n_endorsees, n_newbies, exp_meetups,
```
brenzi, line 1782: oh, I love this!!!!

clangenb: Yes, this is very cool!

PR #19, opened 2021-01-20
Rename currencies to communities

clangenb:
* basically replaced all occurrences of currency/-ies with community/-ies
* rustfmt on save kicked in
* tests in every pallet pass


brenzi (approved):
LGTM

Issue #15, opened 2020-12-22
implement SybilGate pallet with XCMP


brenzi:
serve as XCMP api for other parachains to query digital personhood confidence 

parachain A: encointer/encointer-parachain
parachain B: some test parachain using the new encointer XCMP pallet

subject: individual attendee participating in encointer cermeonies and attempting some sybil-sensitive action on parachain B
issuer: encointer chain
verifier: parachain B

verifier request (from parachain B):
* confidence threshold N of M last ceremonies
* ProofOfAttendance for all attended ceremonies (supplied by subject to verifier, also stating cid's)

issuer response:
* approve/deny

# test setup
[launch local relay-chain - parachain setup](https://github.com/paritytech/cumulus#launch-a-local-setup-including-a-relay-chain-and-a-parachain)

use encointer/encointer-parachain and modify calls accordingly:

in paritytech/polkadot
```
./target/release/polkadot build-spec --chain rococo-local --disable-default-bootnode --raw > rococo-local-cfde-real-overseer.json
./target/release/polkadot --chain rococo-local-cfde-real-overseer.json --alice --tmp --port 30335 --ws-port 9979
# next terminal
./target/release/polkadot --chain rococo-local-cfde-real-overseer.json --bob --tmp --port 30336 --ws-port 9980
```
in encointer/encointer-parachain (parachain A)
```
./target/release/encointer-collator --collator --tmp --parachain-id 1862 --chain encointer-rococo --port 40335 --ws-port 9946 -- --execution wasm --chain ../polkadot/rococo-local-cfde-real-overseer.json --port 30337 --ws-port 9981
# second collator optional
./target/release/encointer-collator --collator --tmp --chain encointer-rococo --parachain-id 1862 --port 40336 --ws-port 9947 -- --execution wasm --chain ../polkadot/rococo-local-cfde-real-overseer.json --port 30338 --ws-port 9982
```
please note: 
  * CLI arguments after the `--` regard the polkadot node spawned by the collator
  * use `rococo-v1` branch of polkadot, substrate, cumulus
  * rebase our collator on paritytech/cumulus `rococo-v1` branch regularly

more resources:
* https://substrate.dev/cumulus-workshop/#/ (currently outdated)
* up to date rococo-v1 instructions https://wiki.polkadot.network/docs/en/build-parachains-rococo#rococo-v1-parachain-requirements

clangenb:
These pallets seem to use xcmp:
[...]

PR #20, opened 2021-02-04
some code restructuring + sybil gate WIP

clangenb:
* Move types introduced by pallets to dedicated `primitives` crate
* Added test-setup crate, which essentially provides macros to decorate the test runtimes used in the pallet tests

* Added a WIP of the `sybil-gate` implementation.  #15


brenzi:
test for sybil gate fails. That would not go well with CI - if we had it. Can you just remove the sybil-gate crate or make the unit tests pass?

clangenb:
> test for sybil gate fails. That would not go well with CI - if we had it. Can you just remove the sybil-gate crate or make the unit tests pass?

Ah yes, sorry forgot about that test. Fixed it

brenzi (approved):
good stuff!

PR #21, opened 2021-02-09
Implement sybil gate

clangenb:
* Finish SybilGate design. Made a successful roundtrip on rococo-v1 with empty `ProofOfPersonhoodRequest`.

**Todo**: SybilProofIssuer needs testing

Other Stuff:
* use `MultiSignature` in tests
* Depend on substrate master branch instead of specific revision to be able to align substrate second level dependencies with polkadot's substrate deps. It turns out that specifying revisions is fairly easy via search and replace in Cargo.lock


brenzi (changes requested):
I can't accept `branch = master`. This messes up the build of encointer-node and is maintenance-heavy.

Please
1. remove the dependency change and leave the substrate rev as is.
2. use cargo patch to change the revision when using pallets for encointer-parachain

can that work?

clangenb:
Well, the core problem is: Polkadot depends on substrate master. Cargo does not recognize that `rev = ...` is compatible with `branch = master`. Even with a Cargo.lock that already has that specific substrate dependency needed by Polkadot. So we can do:

1. Patch, but we would need to fork substrate (or polkadot), as patched dependencies can only be patched by a different registry. So this is an effort too.

2. In this repo, the only way to compile successfully without patching was: 
    1. Making said change to `branch = master`. 
    2. Start compiling, break after dependency resolution
    3. Search&replace revisions in the Cargo.lock for the substrate and Polkadot dependencies
    4. Compile again.

My guess is that is still easier to work with my change to `branch = master`, which would imply that the same change needs to be done in encointer node, otherwise we have the same problem that Cargo does not recognize compatibility the pallets' substrate dependencies and the encointer-node's.

3. Remove sybil gate pallets from the workspace considering they are only  in needed when the encointer node is used as a parachain.

brenzi:
Option 4: split repos and put XCM stuff in encointer/sybil-gate.git

clangenb:
I have been thinking about option 4. Even though I don't like the pallets being scattered, I believe this is the best option from a user's perspective, as Cargo does download and cache the entire git registry from dependencies. This means all the other pallets would be downloaded and cached as well, which is not necessary for the sybil-gate.

Hence, moving it to a separate repository, would reduce the amount of downloaded data.

brenzi:
once cumulus has stabilized, we might move the pallets together again, but I think it is worth moving the XCM stuff to its own repo. Actually, that repo could even be encointer/encointer-parachain itself for dev convenience: Option 5!!!

primitives/src/sybil.rs
```diff
@@ -1,34 +1,100 @@
...
 use crate::ceremonies::ProofOfAttendance;
-use crate::communities::CommunityIdentifier;
 use crate::scheduler::CeremonyIndexType;
 
-pub type ProofOfPersonhoodRequest<Signature, AccountId> =
-    Vec<(CommunityIdentifier, ProofOfAttendance<Signature, AccountId>)>;
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Default, RuntimeDebug)]
+pub struct IssueProofOfPersonhoodConfidenceCall<Signature, AccountId> {
+    call_index: [u8; 2],
+    requester: AccountId,
```
brenzi, line 12: is this the accountid the subject uses on the faucet parachain? why is that needed?
encointer advises users to use a different accountid for each ceremony for improved privacy, so all proofs can have different accountid. And we would also advise to use yet another accountid on the faucet parachain.
The question is: why would the encointer chain need to know this accountid at all?

primitives/src/sybil.rs
```diff
@@ -1,34 +1,100 @@
...
+    request: Vec<ProofOfAttendance<Signature, AccountId>>,
```
brenzi, line 13: ProofOfAttendance should be opaque to the faucet parachain. Please replace this with a `<Vec<u8>`
For cantillon, the payload will be encrypted anyway

primitives/src/sybil.rs
```diff
@@ -1,34 +1,100 @@
...
     attested: CeremonyIndexType,
     last_n_ceremonies: CeremonyIndexType,
+    proofs: Vec<H256>,
 }
 
 impl ProofOfPersonhoodConfidence {
-    pub fn new(attested: CeremonyIndexType, last_n_ceremonies: CeremonyIndexType) -> Self {
+    pub fn new(
+        attested: CeremonyIndexType,
+        last_n_ceremonies: CeremonyIndexType,
```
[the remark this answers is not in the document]
brenzi: the return value for the facuet chain can be rounded to a u32 with full scale eq 1.0, no need to depend on fixed for this

future: communities will have a certain confidence rating too, so I suggest that we already introduce \beta as the confidence of the community as such, which auto-initializes to some very small positive number when the community is registered (zero is bad for testing as the confidence will always be zero)
so, the confidence formula becomes

confidence = sum(\beta * exp(T*\alpha)) | for all attended ceremonies with a valid proof

sybil/gate/src/lib.rs
```diff
@@ -43,77 +51,146 @@ pub trait Config: frame_system::Config {
...
             }
         }
 
         #[weight = 5_000_000]
-        fn set_proof_of_personhood_confidence(
+        /// ### Faucet
+        ///
+        /// Faucet that funds accounts. Currently, this can only be called from other parachains, as
+        /// the ProofOfPersonhood can otherwise not be verified.
+        fn faucet(
```
brenzi, line 145: a faucet does not belong into this pallet. it should have its own pallet `sybil-gate-template`

brenzi: actually, we could leave it, but rename this entire pallet to `sybil-gate-template` to make sure this code is just an example.

brenzi (changes requested):
very nice already. but I think we need to go a step further so we can really tell the story how sybil defense actually works

primitives/src/sybil.rs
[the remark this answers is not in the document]
clangenb: I must say, I like the idea of the exponential function.

However, your observation is not correct (the naming is arguably bad). But I do allow gaps. `last_n_ceremonies` stands for the earliest ceremony, a proof was included, and `attested` is the amount of those `last_n_ceremonies`, which a proof was supplied for.

Hence, the idea was that the recipient of the proof has two pieces of information:
* the earliest ceremony that was attended (current_ceremony_index -`n`)
* how many of the last `n` ceremonies were attended.

I thought that this is even more flexible than an exponentially decaying weight of attended ceremonies.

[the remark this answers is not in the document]
brenzi: Ah, I missed that. Indeed, what you did is sufficient for now. I suggest to open a new issue for the exponential stuff because it will later allow the weighting of community web of trust as well. I'll just copy-paste the idea into an issue.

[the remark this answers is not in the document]
brenzi: https://github.com/encointer/pallets/issues/24

brenzi:
On the matter of naming: 

`sybil/proof-issuer` is not actually issuing proofs. What it issues is a confidence metric for the uniqueness of a subjects digital personhood and it does this by verifying a set of `ProofOfAttendance`. In the paper we called it a uPoP (unique-proof-of-personhood), but that name doesn't really apply to a scalar confidence value we're requesting here. 
I suggest to rename this crate to `personhood-uniqueness-verifier` and put it in the root folder directly. *uniqueness* is kind of a scalar value representing the certainty of uniqueness.
This crate is the *server* supplying the service of verifying a uPoP if you so wish. It should also expose an RPC method for non-parachain parties to verify uniqueness too (please draft an issue for this task after renaming) This feature is not only useful for XCM!

`sybil/gate` should be in the root folder too and renamed to `sybil-gate-template` as suggested in a comment above. We might want to put this in its own repo later for easy forking for downstream projects but now its probably better to leave it here

primitives/src/sybil.rs
clangenb: Yes, this is absolutely unnecessary, this has only been used as the key for the `PendingRequests`, which is a bad choice.

This could hash of the ProofOfPersonHood request, which makes more sense.

primitives/src/sybil.rs
clangenb: agreed

sybil/gate/src/lib.rs
clangenb: Yes, I think it is better to leave it there, but it should be more obvious that this is an example.

brenzi: agreed, see naming comment below

clangenb:
> On the matter of naming:
> 
> `sybil/proof-issuer` is not actually issuing proofs. What it issues is a confidence metric for the uniqueness of a subjects digital personhood and it does this by verifying a set of `ProofOfAttendance`. In the paper we called it a uPoP (unique-proof-of-personhood), but that name doesn't really apply to a scalar confidence value we're requesting here.
> I suggest to rename this crate to `personhood-uniqueness-verifier` and put it in the root folder directly. _uniqueness_ is kind of a scalar value representing the certainty of uniqueness.
> This crate is the _server_ supplying the service of verifying a uPoP if you so wish. It should also expose an RPC method for non-parachain parties to verify uniqueness too (please draft an issue for this task after renaming) This feature is not only useful for XCM!
> 
> `sybil/gate` should be in the root folder too and renamed to `sybil-gate-template` as suggested in a comment above. We might want to put this in its own repo later for easy forking for downstream projects but now its probably better to leave it here

I was also not satisfied with the naming, but simply could not think of something better yet.

Your suggestion is already better, but I don't think the pallet _verifies_ uniqueness but rather returns an _estimate_, _measure_ or _rating._

Additionally, shall we keep the crate's name more generic, i.e. `personhood-uniqueness` or simply `personhood` in case we might add more calls revolving around personhood in the future?

clangenb:
> sybil/gate should be in the root folder too and renamed to sybil-gate-template as suggested in a comment above. We might want to put this in its own repo later for easy forking for downstream projects but now its probably better to leave it here

Here, I agree with both. I think we should move it as soon as we finalized the first version.

brenzi:
what about `personhood-oracle`? It actually provides a kind of oracle service

clangenb:
> what about `personhood-oracle`? It actually provides a kind of oracle service

Oh, I like this!

Issue #28, opened 2021-02-26
enable community customization and branding


brenzi:
Currently, our chain spec only allows to set bootstrappers, but name is ignored:

```
    "community_meta": {
        "name": "mediterranean test community",
        "bootstrappers": [
          "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
          "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
          "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
          "5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy",
          "5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw",
          "5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL"
        ]
    },
```

Moving towards mainnet, we want to allow branding. The app shall no longer show the human-unfriendly `CommunityIdentifier` and we want to allow customization

1. `name`: a long name, like "Züri Leu"
2. `symbol` abbreviation, like "LEU"
3. `icons` a multi-resolution resource for the community icon
4. `theme` an optional color scheme or other customizable styles to shape app appearance
4. `url` a optional link to a community site
4. `demurrage_per_block`:  to override the default
5. `nominal_income`: the amount of UBI to be paid out for every attended ceremony. to override the default

For icons, we could use the favicon format, like manifest.json
```
  "icons": [{
        "src": "<ipfs#>",
        "sizes": "64x64",
      }, {
        "src": "<ipfs#>",
        "sizes": "64x64"
      }, {
        "src": "<ipfs#>",
        "sizes": "128x128",
        "density": 2
      }],
```

TODO:
1. extend extrinsic call to include above customizations
2. offer sudo calls to change each field
3. override `nominal_income` in `pallet-encointer-ceremonies` and `demurrage_per_block` in `pallet-encointer-balances`
3. modify encointer/explorer to respect and show the new customizations (open new issue in other repo)
4. modify encointer/encointer-wallet-flutter to respect and show the new customizations (open new issue in other repo)

see `bazaar`for IPFS integration

clangenb:
[...]

Issue #9, opened 2020-10-17
currencies limited to 1000 locations


brenzi:
With the current implementation it is not possible to register more than about 1000 locations for a currency, this would limit the population size to 10'000 which is not acceptable.

When trying to register 2000 locations, I get
```
Oct 17 23:41:11.999  INFO ⌛️ Discarding proposal for slot 267161810; block production took too long                                                                                 
Oct 17 23:41:11.999  WARN Encountered consensus error: ClientImport("Timeout in the Slots proposer")                                                                                
```

This should not surprise, as the validation of locations and their respective distances has complexity O(n^2)

We should
1. limit the size of the location vec in the call to ~ 100
2. allow to add more locations over time, 100 at the time (but the complexity is still O(M * N^2) where M is the total number of locations for all currencies
3. offchain worker in TEE?

PR #31, opened 2021-03-16
Enable community branding

### LINKED ISSUE #14: constrain dispatchable argument lengths
currencies uses `Vec<Location>` and `Vec<AccountId>` but doesn't limit the size of these neither makes the weight dependant on it.

We should do both!

clangenb:
-  Draft implementation of #28 
    - extend `new_community` extrinsic to include `CommunityMetadata` and `Demurrage`/`NominalIncome` overrides.
    - add sudo calls to override the above settings given in the `new_community` extrinsic

- add generic `register_test_currency` in `test-utils` to remove redundant definitions of it.
- constrain amount of bootstrapers. Closes #14
- some other small fixes.

### DIFF
--- a/communities/src/lib.rs
+++ b/communities/src/lib.rs
@@ -131,8 +131,8 @@ decl_module! {
             validate_demurrage(&demurrage).map_err(|_| <Error<T>>::InvalidDemurrage)?;
             Self::ensure_cid_exists(&cid)?;
 
-            <DemurragePerBlock>::insert(&cid, demurrage);
-            Self::deposit_event(RawEvent::DemurrageUpdated(cid));
+            <DemurragePerBlock>::insert(&cid, &demurrage);
+            Self::deposit_event(RawEvent::DemurrageUpdated(cid, demurrage.to_bits()));
             debug::info!(target: LOG, " updated demurrage for cid: {:?}", cid);
         }
 
@@ -143,8 +143,8 @@ decl_module! {
             validate_nominal_income(&nominal_income).map_err(|_| <Error<T>>::InvalidDemurrage)?;
             Self::ensure_cid_exists(&cid)?;
 
-            <NominalIncome>::insert(&cid, nominal_income);
-            Self::deposit_event(RawEvent::NominalIncomeUpdated(cid));
+            <NominalIncome>::insert(&cid, &nominal_income);
+            Self::deposit_event(RawEvent::NominalIncomeUpdated(cid, nominal_income.to_bits()));
             debug::info!(target: LOG, " updated nominal income for cid: {:?}", cid);
         }
     }
@@ -157,12 +157,12 @@ decl_event!(
     {
         /// A new community was registered \[who, community_identifier\]
         CommunityRegistered(AccountId, CommunityIdentifier),
-        /// CommunityMetadata was updated \[who, community_identifier\]
+        /// CommunityMetadata was updated \[community_identifier\]
         MetadataUpdated(CommunityIdentifier),
-        /// A community's nominal income was updated \[who, community_identifier\]
-        NominalIncomeUpdated(CommunityIdentifier),
-        /// A community's demurrage was updated \[who, community_identifier\]
-        DemurrageUpdated(CommunityIdentifier),
+        /// A community's nominal income was updated \[community_identifier, new_income\]
+        NominalIncomeUpdated(CommunityIdentifier, i128),
+        /// A community's demurrage was updated \[community_identifier, new_demurrage\]
+        DemurrageUpdated(CommunityIdentifier, i128),
     }
 );
 
--- a/communities/src/lib.rs
+++ b/communities/src/lib.rs
@@ -28,7 +28,7 @@ use frame_support::{
     debug, decl_error, decl_event, decl_module, decl_storage, ensure,
     storage::{StorageMap, StorageValue},
 };
-use frame_system::ensure_signed;
+use frame_system::{ensure_root, ensure_signed};
 
 use rstd::prelude::*;
 
@@ -141,20 +141,16 @@ decl_module! {
         #[weight = 10_000]
         fn update_community_medadata(origin, cid: CommunityIdentifier, community_metadata: CommunityMetadataType) {
             debug::RuntimeLogger::init();
-            let sender = ensure_signed(origin)?;
+            ensure_root(origin)?;
 
             if !Self::community_identifiers().contains(&cid) {
                 return Err(<Error<T>>::CommunityInexistent)?;
             }
 
-            if !<Bootstrappers<T>>::get(cid).contains(&sender) {
-                return Err(<Error<T>>::MissingPrivileges)?;
-            }
-
             // Todo: Validate metadata??
             <CommunityMetadata>::insert(&cid, community_metadata);
-            Self::deposit_event(RawEvent::CommunityMetadataUpdated(sender, cid));
-            debug::info!(target: LOG, "registered community with cid: {:?}", cid);
+            Self::deposit_event(RawEvent::CommunityMetadataUpdated(cid));
+            debug::info!(target: LOG, "updated community metadata for cid: {:?}", cid);
         }
     }
 }
@@ -167,8 +163,7 @@ decl_event!(
         /// A new community was registered \[who, community_identifier\]
         CommunityRegistered(AccountId, CommunityIdentifier),
         /// CommunityMetadata was updated \[who, community_identifier\]
-        CommunityMetadataUpdated(AccountId, CommunityIdentifier),
-    }
+        CommunityMetadataUpdated(CommunityIdentifier),
 );
 
 decl_error! {
@@ -179,8 +174,6 @@ decl_error! {
         MinimumDistanceViolationToDateLine,
         /// minimum distance violated towards other community's location
         MinimumDistanceViolationToOtherCommunity,
-        /// Only bootstrappers can call this function
-        MissingPrivileges,
         /// Can't register community that already exists
         CommunityAlreadyRegistered,
         /// Community does not exist yet
--- a/communities/src/lib.rs
+++ b/communities/src/lib.rs
@@ -89,7 +89,7 @@ decl_module! {
                 validate_demurrage(&d).map_err(|_| <Error<T>>::InvalidDemurrage)?;
             }
             if let Some(i) = nominal_income {
-                validate_nominal_income(&i).map_err(|_| <Error<T>>::InvalidDemurrage)?;
+                validate_nominal_income(&i).map_err(|_| <Error<T>>::InvalidNominalIncome)?;
             }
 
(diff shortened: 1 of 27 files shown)


clangenb:
@brenzi You might want to have a look at this draft? In fact, would say it is almost finished.

The only todos that are left concern the exact typedefs for the `Favicon` and the `Theme`, but I want to check and implement on the app side first, to see which serialization format works.

clangenb:
Actually, the above is not quite true. I think the types should work just fine and are good as a starting point.

communities/src/lib.rs
```diff
@@ -53,7 +57,11 @@ decl_storage! {
         Locations get(fn locations): map hasher(blake2_128_concat) CommunityIdentifier => Vec<Location>;
         Bootstrappers get(fn bootstrappers): map hasher(blake2_128_concat) CommunityIdentifier => Vec<T::AccountId>;
         CommunityIdentifiers get(fn community_identifiers): Vec<CommunityIdentifier>;
-        CommunityProperties get(fn community_properties): map hasher(blake2_128_concat) CommunityIdentifier => CommunityPropertiesType;
+        CommunityMetadata get(fn community_metadata): map hasher(blake2_128_concat) CommunityIdentifier => CommunityMetadataType;
+        /// If it is empty, the genesis config's default is used.
```
brenzi, line 61: as you set the default in the consuming pallet's genesis, not in this pallet's genesis, the comment is misleading

communities/src/lib.rs
```diff
@@ -67,58 +75,77 @@ decl_module! {
         // where n is the number of all locations of all communities
         // this should be run off-chain in substraTEE-worker later
         #[weight = 10_000]
-        pub fn new_community(origin, loc: Vec<Location>, bootstrappers: Vec<T::AccountId>) -> DispatchResult {
+        pub fn new_community(
+            origin,
+            loc: Vec<Location>,
+            bootstrappers: Vec<T::AccountId>,
```
brenzi, line 81: As you're touching the code here, I'd suggest to resolve #14 at the same time.
Every `vec` argument should be constrained in size, i.e. with `ensure` at the beginning of the fn

Locations: no more than 1000 for now (because of #9)
bootstrappers: no more than 12

primitives/src/common.rs
```diff
@@ -0,0 +1,14 @@
+use rstd::vec::Vec;
+
+/// Substrate runtimes provide no string type. Hence, for arbitrary data of varying length the
+/// `Vec<u8>` is used. In the polkadot-js the typedef `Text` is used to automatically
+/// utf8 decode bytes into a string.
+pub type PalletString = Vec<u8>;
```
brenzi, line 6: why rename the original type?

communities/src/lib.rs
```diff
@@ -67,58 +75,77 @@ decl_module! {
...
-                        }
-                    }
-                }
-            }
+            // All checks done, now mutate state
 
             <CommunityIdentifiers>::mutate(|v| v.push(cid));
             <Locations>::insert(&cid, &loc);
             <Bootstrappers<T>>::insert(&cid, &bootstrappers);
+            // Todo: Validate metadata??
```
brenzi, line 105: yes!
* validate that contents is utf8 if need be
* limit length of PalletStrings (at beginning of fn, not at the end)

communities/src/lib.rs
```diff
@@ -67,58 +75,77 @@ decl_module! {
...
-            Ok(())
+        }
+
+        #[weight = 10_000]
+        fn update_community_medadata(origin, cid: CommunityIdentifier, community_metadata: CommunityMetadataType) {
+            debug::RuntimeLogger::init();
+            ensure_root(origin)?;
+            Self::ensure_cid_exists(&cid)?;
+
+            // Todo: Validate metadata??
```
brenzi, line 121: dito

communities/src/lib.rs
```diff
@@ -128,7 +155,14 @@ decl_event!(
     where
         AccountId = <T as frame_system::Config>::AccountId,
     {
+        /// A new community was registered \[who, community_identifier\]
         CommunityRegistered(AccountId, CommunityIdentifier),
+        /// CommunityMetadata was updated \[who, community_identifier\]
+        MetadataUpdated(CommunityIdentifier),
+        /// A community's nominal income was updated \[who, community_identifier\]
+        NominalIncomeUpdated(CommunityIdentifier),
```
brenzi, line 163: it would be lightweight to add the new value here

communities/src/lib.rs
```diff
@@ -128,7 +155,14 @@ decl_event!(
...
+        /// A community's demurrage was updated \[who, community_identifier\]
+        DemurrageUpdated(CommunityIdentifier),
```
brenzi, line 165: and here

primitives/src/communities.rs
```diff
@@ -3,14 +3,38 @@ use fixed::types::I64F64;
...
+///
+/// Todo: Other sanity checks, e.g., 0 < e^(demurrage_per_block*sum(phase_durations)) < 1?
+pub fn validate_demurrage(demurrage: &Demurrage) -> Result<(), ()> {
+    if demurrage < &Demurrage::from_num(0) {
+        return Err(());
+    }
+    Ok(())
+}
+
+/// Ensure that the demurrage is in a sane range.
```
brenzi, line 30: copy-paste comment wrong

primitives/src/communities.rs
```diff
@@ -19,9 +43,38 @@ pub struct Location {
...
+    /// optional link to a community site
+    pub url: Option<PalletString>,
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Default, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct Favicon {
+    src: IpfsCid,
+    /// e.g. "64x64", "128x128", ...
+    sizes: PalletString,
```
brenzi, line 65: why is sizes plural if this type defines a single image resource?

primitives/src/communities.rs
```diff
@@ -19,9 +43,38 @@ pub struct Location {
...
-    pub name_utf8: Vec<u8>,
-    pub demurrage_per_block: Demurrage,
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct CommunityMetadata {
+    /// utf8 encoded name
+    pub name: PalletString,
+    /// utf8 encoded abbreviation of the name
+    pub symbol: PalletString,
+    /// multi-resolution resource for the community icon
+    pub icons: Vec<Favicon>,
```
brenzi, line 53: I'm undecided if we should really bloat the chain with so many ipfs cid's. we could link to a single json on ipfs defining the favicons

brenzi: yeah, let's do a single cid. there is no added value in having this all onchain. immutability is given anyway

brenzi (changes requested):
I like this.

For future PR's: Please try keeping them small and make one PR per change to be reviewed. Here I had to review test improvements and cosmetics along with the actual task, which makes it confusing

primitives/src/common.rs
clangenb: I wanted to make it semantically clear that the data underlying the `Vec<u8>` is an utf8 encoded string in cases where it is not obvious.

For instance, we previously had the field `name_utf8` in the `CommunityProperties`. Here we could safely remove the suffix without potentially causing a misunderstanding.

I believe, trying to be as explicit as possible gives better understanding when reading the code.

communities/src/lib.rs
clangenb: agreed

communities/src/lib.rs
clangenb: > For future PR's: Please try keeping them small and make one PR per change to be reviewed. Here I had to review test improvements and cosmetics along with the actual task, which makes it confusing

Yes sorry about that. I agree and I should have taken the extra effort. It kind of emerged organically, as I had to use the `register_test_currency` in a new place and thought that instead of copy-pasting it to yet another location I solve the issue. But I will be extra strict with myself the next time. :)

primitives/src/communities.rs
clangenb: > For future PR's: Please try keeping them small and make one PR per change to be reviewed. Here I had to review test improvements and cosmetics along with the actual task, which makes it confusing

Yes sorry about that. I agree and I should have taken the extra effort. It kind of emerged organically, as I had to use the `register_test_currency` in a new place and thought that instead of copy-pasting it to yet another location I solve the issue. But I will be extra strict with myself the next time. :)

clangenb:
> For future PR's: Please try keeping them small and make one PR per change to be reviewed. Here I had to review test improvements and cosmetics along with the actual task, which makes it confusing

Yes sorry about that. I agree and I should have taken the extra effort. It kind of emerged organically, as I had to use the `register_test_currency` in a new place and thought that instead of copy-pasting it to yet another location I solve the issue. But I will be extra strict with myself the next time. :)

primitives/src/communities.rs
clangenb: typo, but it gets removed anyway now.

communities/src/lib.rs
clangenb: Of what value are you speaking?

clangenb: ah ok, I got it now!

communities/src/lib.rs
```diff
@@ -157,17 +157,27 @@ decl_event!(
...
         /// A new community was registered \[who, community_identifier\]
         CommunityRegistered(AccountId, CommunityIdentifier),
-        /// CommunityMetadata was updated \[who, community_identifier\]
+        /// CommunityMetadata was updated \[community_identifier\]
         MetadataUpdated(CommunityIdentifier),
-        /// A community's nominal income was updated \[who, community_identifier\]
-        NominalIncomeUpdated(CommunityIdentifier),
-        /// A community's demurrage was updated \[who, community_identifier\]
-        DemurrageUpdated(CommunityIdentifier),
+        /// A community's nominal income was updated \[community_identifier, new_income\]
```
clangenb, line 162: Here I had to make a small workaround due to `substrate-fixed` not implementing some traits. The value is displayed as the result from `nominal_income.to_bits()`.

brenzi: ok. what traits?

clangenb: This turned out to be an error on my side.

I put `NominalIncome` (storage) instead of `NominalIncometType` in the event, which gave the trait errors. Putting the correct type in the event solved the issue.

communities/src/lib.rs
clangenb: Found a `no_std` compatible utf8 encoding lib [encoiding_rs](https://github.com/hsivonen/encoding_rs)

Will add validation with it.

brenzi:
as you are changing the API, please bump all pallet versions to 0.4.0

clangenb:
ready for another review

brenzi (approved):

PR #37, opened 2021-04-29
Add custom rpc returning all community names, using offchain-indexing.

clangenb:
* use offchain-indexing to store all the communities' names in offchain storage
* custom rpc assembles the individual names and returns: `Vec<CidName>` (== `Vec<(CommunityIdentifier, Name)>`)
* The rpc caches the `Vec<CidName>`, for subsequent calls
* The pallet invalidates the cache when a new community is added or a community's name changed.


brenzi (commented):
It would be nice to see more testing. I only see a cache test.
How can we test that the rpc interface works and returns the desired content?
Do we need an end2end test including our CLI client?

clangenb:
Yes, this is as much as we can test without instantiation of a runtime and a client, as this calls into the runtime:

https://github.com/encointer/pallets/blob/9202a3eeadd072c5ff540a45f02a143f649d213f/communities/rpc/src/lib.rs#L112

I have looked into substrates' [test-runtime](https://github.com/paritytech/substrate/tree/master/test-utils/runtime) and the corresponding client, but I deemed it an overkill.

For easy testing I have adjusted the encointer clients `list-communities` to display the results of this rpc.

clangenb:
What do you think about the name `CidName`? I am not very satisfied with it...

Shall we simply name it Community?

brenzi:
most unambiguous would be `CommunityName` I think

primitives/src/communities.rs
```diff
@@ -56,6 +56,19 @@ pub struct CommunityMetadata {
     pub url: Option<PalletString>,
 }
 
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct CidName {
+    pub cid: CommunityIdentifier,
+    pub name: PalletString,
+}
+
```
clangenb, line 65: Hmm, I don't quite agree that `CommunityName` would be a good name. I'd prefer `Community` as it is more than the name, but less than its metadata.

But `CidName` is awful. It hurts everytime I read it. XD

brenzi: ah, it's a double-identifier, I see. Then I think `CidName` is actually quite precise. No hurt in my eyes ;-)

brenzi: but `Community` is ok with me too. I don't want to make you suffer

PR #46, opened 2021-06-11
Bazaar Refactoring

NicJak:
The storage model of bazaar is reduced to two storage_double_maps and ready for usage with anonymous proxies

this lays the foundation for https://github.com/encointer/encointer-wallet-flutter/issues/122


bazaar/README.md
```diff
@@ -41,14 +41,89 @@ optional:
...
+   * businesses per community
+   * articles per community
+   * articles per businesses
+
+#### Proposed Storage Model
 
 ```
-BusinessRegistry: double_map(cid, AccountId) -> (url, block_number)
-OfferingRegistry: double_map(cid, AccountId) -> Vec<(url, block_number)>
+BusinessRegistry: double_map(CommunityId, BusinessAccountId) -> (business_url, last_oid, block_number)
```
clangenb, line 62: I think we decided to get rid of the `block_number`

bazaar/README.md
```diff
@@ -41,14 +41,89 @@ optional:
...
+   * create business
+   * edit business
+   * delete business
+2. Offering Management
+   * create offering
+   * edit offering
+   * delete offering
+3. View (RPC)
+   * businesses per community
+   * articles per community
```
clangenb, line 56: I would consistently use offerings as names instead of articles. All storage fields in a pallet are implicitly exposed via RPC and will therefore follow the storage naming anyhow.

bazaar/src/lib.rs
```diff
@@ -89,73 +88,113 @@ decl_module! {
...
+
+            Ok(())
+        }
+
+        #[weight = 10_000]
+        pub fn create_offering(origin, cid: CommunityIdentifier, url: PalletString) -> DispatchResult {
+            // Check that the extrinsic was signed and get the signer
+            let sender = ensure_signed(origin)?;
+            // Check that the supplied community is actually registered
+            ensure!(<encointer_communities::Module<T>>::community_identifiers().contains(&cid),
```
clangenb, line 148: This check is actually redundant. The next ensure will fail anyhow if the cid does not exist. The same goes for the other methods.

bazaar/src/lib.rs
```diff
@@ -89,73 +88,113 @@ decl_module! {
...
+        pub fn update_offering(origin, cid: CommunityIdentifier, oid: OfferingIdentifier, url: PalletString) -> DispatchResult {
+            // Check that the extrinsic was signed and get the signer
+            let sender = ensure_signed(origin)?;
+            // Check that the supplied community is actually registered
+            ensure!(<encointer_communities::Module<T>>::community_identifiers().contains(&cid),
+                Error::<T>::InexistentCommunity);
+
+            let business_identifier = BusinessIdentifier{community_identifier: cid, business_account: sender.clone()};
+
+            ensure!(OfferingRegistry::<T>::contains_key(business_identifier.clone(), oid.clone()), Error::<T>::InexistentOffering);
```
clangenb, line 172: I am not 100% sure, but I believe you can always pass borrows as keys in the map instead of clone.

bazaar/src/lib.rs
```diff
@@ -89,73 +88,113 @@ decl_module! {
...
+
+        #[weight = 10_000]
+        pub fn delete_offering(origin, cid: CommunityIdentifier, oid: OfferingIdentifier) -> DispatchResult {
+            // Check that the extrinsic was signed and get the signer
+            let sender = ensure_signed(origin)?;
+            // Check that the supplied community is actually registered
+            ensure!(<encointer_communities::Module<T>>::community_identifiers().contains(&cid),
+                Error::<T>::InexistentCommunity);
+
+            let business_identifier = BusinessIdentifier{community_identifier: cid, business_account: sender.clone()};
```
clangenb, line 189: Don't you think this would be a less verbose if you implemented the `new` constructor?

You could call it without the field annotations.

bazaar/src/tests.rs
```diff
@@ -53,261 +51,225 @@ impl ExtBuilder {
...
+        assert!(EncointerBazaar::delete_business(Origin::signed(alice()), cid).is_ok());
+
+        assert_eq!(EncointerBazaar::business_registry(cid, alice()), BusinessData::default());
+        assert_eq!(EncointerBazaar::business_registry(cid, bob()), BusinessData { url: url1(), last_oid: 3 });
     });
 }
 
 #[test]
-fn removal_of_shop_works() {
+fn delete_business_inexistent() {
```
clangenb, line 148: I would prefer if the test names read as a sentence and are as explicit as possible. Here I am not certain if it should fail or not.

I believe that would be better:  `deleting_inexistent_business_is_err` or `deleting_inexistent_business_is_noop`, what do you think about that?

This also applies to the other tests.

clangenb (changes requested):

clangenb (approved):
Looks very nice and clean

bazaar/src/lib.rs
```diff
@@ -47,39 +47,38 @@ pub trait Config: frame_system::Config + encointer_communities::Config {
...
-        // Item owner
-        pub ShopOwner get(fn shop_owner): double_map hasher(blake2_128_concat) CommunityIdentifier, hasher(blake2_128_concat) ShopIdentifier => T::AccountId;
-        pub ArticleOwner get(fn article_owner): double_map hasher(blake2_128_concat) CommunityIdentifier, hasher(blake2_128_concat) ArticleIdentifier => (T::AccountId, ShopIdentifier);
-        // The set of all shops and articles per community
-        pub ShopRegistry get(fn shop_registry): map hasher(blake2_128_concat) CommunityIdentifier => Vec<ShopIdentifier>;
-        pub ArticleRegistry get(fn article_registry): map hasher(blake2_128_concat) CommunityIdentifier => Vec<ArticleIdentifier>;
+        pub BusinessRegistry get(fn business_registry): double_map hasher(blake2_128_concat) CommunityIdentifier, hasher(blake2_128_concat) T::AccountId => BusinessData;
+        pub OfferingRegistry get(fn offering_registry): double_map hasher(blake2_128_concat) BusinessIdentifier<T::AccountId>, hasher(blake2_128_concat) OfferingIdentifier => OfferingData;
```
brenzi, line 51: This is not what we had agreed on in our design thread. The idea was to have the CommunityIdentifier in the primary prefix together with BusinessIdentifier in a tuple

brenzi (changes requested):
I'll review again after fixing storage model (or arguing well against ;-)

bazaar/src/lib.rs
clangenb: Yes I agree, you had it included before. Why was it removed?

NicJak: Maybe I'm misunderstanding the issue, but to me "BusinessAccountId is the public key of an anonymous proxy aka the business. The described tuples/triples are wrapped in value type objects." (copied from readme) indicates, that the tuple is to be implemented using a proper struct ?

NicJak: Businessidentifier is a struct consisting of AcvountId and CommunityIdentifier

clangenb: Ah yes, I forgot about that change, although I agree that the naming is not ideal. 

Structs and tuple are codec-wise identical. However, I generally vote for strongly typed interfaces. So, I favor struct over tuple, as it is as expressive as possible.

primitives/src/bazaar.rs
```diff
@@ -14,7 +14,53 @@
...
-pub type ShopIdentifier = IpfsCid;
-pub type ArticleIdentifier = IpfsCid;
+use crate::common::PalletString;
+use crate::communities::CommunityIdentifier;
+
+#[derive(Encode, Decode, Default, RuntimeDebug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct BusinessIdentifier<AccountId> {
+    pub community_identifier: CommunityIdentifier,
+    pub business_account: AccountId,
```
brenzi, line 29: I suggest `controller_account`

primitives/src/bazaar.rs
```diff
@@ -14,7 +14,53 @@
...
+
+#[derive(Encode, Decode, Default, RuntimeDebug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct BusinessData {
+    pub url: PalletString,
+    pub last_oid: u32,
+}
+
+impl BusinessData {
+    pub fn new(url: PalletString, last_oid: u32) -> BusinessData {
```
brenzi, line 49: shouldn't new BusinessData always be initialized with last_oid = 0?

bazaar/src/lib.rs
brenzi: ah, sorry, should've read type definition properly. I'm not against using a struct here. But as we're only using this as an index into a double map, I don't think it is bad practise to use a tuple.

clangenb: I don't think it really matters either.

primitives/src/bazaar.rs
NicJak: this constructor is mainly used in the tests... they use it to verify, that last_oid does modified, when updates to the business or erroneous calls that should not alter data at all are being called

NicJak: on second thought, this issue might resolve itself when I use the event system to properly do the checks. We will see tomorrow afternoon. :)

bazaar/src/tests.rs
```diff
@@ -53,261 +51,213 @@ impl ExtBuilder {
...
-        let shops = EncointerBazaar::shop_registry(cid);
-        let alices_shops = EncointerBazaar::shops_owned(cid, alice);
-
-        // assert that shop was not added
-        assert_eq!(shops.contains(&alice_shop), false);
-        assert_eq!(alices_shops.contains(&alice_shop), false);
+        let cid = create_cid();
+
+        assert!(EncointerBazaar::create_business(Origin::signed(alice()), cid, url()).is_ok());
+        assert!(EncointerBazaar::create_business(Origin::signed(alice()), cid, url1()).is_err());
```
brenzi, line 102: wouldn't it be safer to test for a specific error? Otherwise we're not sure the correct error has been thrown

bazaar/src/tests.rs
```diff
@@ -53,261 +51,213 @@ impl ExtBuilder {
...
+    ExtBuilder::build().execute_with(|| {
+        let cid = create_cid();
+        BusinessRegistry::<TestRuntime>::insert(cid, alice(), BusinessData::new(url(), 2));
+        BusinessRegistry::<TestRuntime>::insert(cid, bob(), BusinessData::new(url1(), 2));
+        OfferingRegistry::<TestRuntime>::insert(BusinessIdentifier::new(cid, alice()), 1, OfferingData::new(url()));
+        OfferingRegistry::<TestRuntime>::insert(BusinessIdentifier::new(cid, bob()), 1, OfferingData::new(url1()));
+
+        assert!(EncointerBazaar::delete_business(Origin::signed(alice()), cid).is_ok());
+
+        assert_eq!(EncointerBazaar::business_registry(cid, alice()), BusinessData::default());
```
brenzi, line 259: hmm, isn't there a way to test for null specifically?

brenzi (approved):
looks very clean 👍 
just nice to have wishes left. will approve. Let me know if you want to merge

bazaar/src/tests.rs
NicJak: done

bazaar/src/tests.rs
NicJak: we coud check for it by directly acessing the double_storage_map (i.e. accessing internal state) and checking for "contains"

NicJak:
do not merge yet, I still have one question remaining that we could discuss tomorrow..

NicJak:
ok to merge :)

PR #47, opened 2021-06-21
Bazaar rpc

NicJak:
implement bazaar rpcs

### DIFF
--- a/bazaar/rpc/src/lib.rs
+++ b/bazaar/rpc/src/lib.rs
@@ -1,0 +1,106 @@
+// Copyright (c) 2019 Alain Brenzikofer
+// This file is part of Encointer
+//
+// Encointer is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Encointer is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+use jsonrpc_core::{Error, ErrorCode, Result};
+use jsonrpc_derive::rpc;
+use sp_api::{Decode, Encode, ProvideRuntimeApi};
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use std::sync::Arc;
+use sc_rpc::DenyUnsafe;
+
+use encointer_bazaar_rpc_runtime_api::BazaarApi as BazaarRuntimeApi;
+use encointer_primitives::{communities::CommunityIdentifier, bazaar::BusinessIdentifier};
+use encointer_primitives::bazaar::{OfferingData, BusinessData};
+
+#[rpc]
+pub trait BazaarApi<BlockHash, AccountId>
+    where AccountId: 'static + Encode + Decode + Send + Sync, {
+    #[rpc(name = "bazaar_getBusinesses")]
+    fn get_businesses(&self, cid: CommunityIdentifier, at: Option<BlockHash>) -> Result<Vec<BusinessData>>;
+    #[rpc(name = "bazaar_getOfferings")]
+    fn get_offerings(&self, cid: CommunityIdentifier, at: Option<BlockHash>) -> Result<Vec<OfferingData>>;
+    #[rpc(name = "bazaar_getOfferingsForBusiness")]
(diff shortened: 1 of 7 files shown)


bazaar/rpc/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+            .map(|bid| bid.1.clone())
+            .collect());
+    }
+
+    fn get_offerings(&self, cid: CommunityIdentifier, at: Option<<Block as BlockT>::Hash>) -> Result<Vec<OfferingData>> {
+        let api = self.client.runtime_api();
+        let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
+        return Ok(api.get_businesses(&at, &cid)
+            .map_err(runtime_error_into_rpc_err)?.iter()
+            .flat_map(|bid| api.get_offerings(&at, &BusinessIdentifier::new(cid, bid.0)))
```
clangenb, line 74: Nice! I must say I did not know of flat map before!

bazaar/rpc/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::sync::Arc;
+
+use encointer_bazaar_rpc_runtime_api::BazaarApi as BazaarRuntimeApi;
+use encointer_primitives::{communities::CommunityIdentifier, bazaar::BusinessIdentifier};
+use encointer_primitives::bazaar::{OfferingData, BusinessData};
+
+#[rpc]
+pub trait BazaarApi<BlockHash, AccountId>
+    where AccountId: 'static + Encode + Decode + Send + Sync, {
+    #[rpc(name = "bazaar_getOfferings")]
```
clangenb, line 31: Do you intentionally only mark get_businesses with `#[rpc]`? And did you confuse the name?

I have not studied the internals of the macro closely, but I believe you need to add the attribute to each function you want to expose as rpc.

bazaar/src/lib.rs
```diff
@@ -185,3 +185,14 @@ decl_module! {
         }
     }
 }
+
+impl<T: Config> Module<T>
+{
+    pub fn get_businesses(cid: &CommunityIdentifier) -> Vec<(T::AccountId, BusinessData)> {
```
clangenb, line 191: Hmm, I haven't thought about this before: We iterate through the whole storage here. I think we should either mark the rpc as unsafe, i.e., only expose the rpc if the node is started with `--unsafe-rpc-external` (exact flag name might be different), or we cache the Vec offchain, similar to how we do it in the communities api. Such that subsequent calls will be less expensive.

I don't know about the latter though because caching is only useful, if we don't mark it dirty too often...

Maybe use `DenyUnsafe`?`

clangenb (changes requested):
Except for my two remarks, very nice and clean!

bazaar/src/lib.rs
NicJak: I am unsure about the effects of such unsafe markings. Caching businesses might be discussable, doing it for offerings might be too optimistic.

clangenb: The effect is minimal, it does nothing besides not accepting the rpc, when the unsafe flag isn't set.

bazaar/rpc/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let api = self.client.runtime_api();
+        let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
+        return Ok(api.get_businesses(&at, &cid)
+            .map_err(runtime_error_into_rpc_err)?.iter()
+            .map(|bid| bid.1.clone())
+            .collect());
+    }
+
+    fn get_offerings(&self, cid: CommunityIdentifier, at: Option<<Block as BlockT>::Hash>) -> Result<Vec<OfferingData>> {
```
clangenb, line 78: `check_if_safe()?;` (note the question mark) implicitly performs the `Error::from(error)` conversion.

clangenb (changes requested):
Tiny syntactic sugar remark

clangenb:
Looks good to me. Seems ready to merge?

clangenb (approved):
Oops, I forgot to approve and add brenzi to review. Sorry for stalling this.

LGTM.

bazaar/rpc/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+        Client: Send + Sync + 'static + ProvideRuntimeApi<Block> + HeaderBackend<Block>,
+        Client::Api: BazaarRuntimeApi<Block, AccountId> {
+    fn get_businesses(&self, cid: CommunityIdentifier, at: Option<<Block as BlockT>::Hash>) -> Result<Vec<BusinessData>> {
+        self.deny_unsafe.check_if_safe()?;
+
+        let api = self.client.runtime_api();
+        let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
+        return Ok(api.get_businesses(&at, &cid)
+            .map_err(runtime_error_into_rpc_err)?.iter()
+            .map(|bid| bid.1.clone())
```
clangenb, line 72: Just seen this now. Are you sure you need the `clone()`?

PR #48, opened 2021-07-20
Anonymous proxy

NicJak:
use anonymous proxies to control businesses

### DIFF
--- a/bazaar/src/lib.rs
+++ b/bazaar/src/lib.rs
@@ -25,6 +25,7 @@
 
 mod mock;
 mod tests;
+
 #[cfg(test)]
 extern crate approx;
 
@@ -32,7 +33,7 @@ use frame_support::{
     decl_error, decl_event, decl_module, decl_storage,
     dispatch::DispatchResult,
     ensure,
-    storage::{StorageDoubleMap},
+    storage::{StorageDoubleMap, IterableStorageDoubleMap},
 };
 use frame_system::ensure_signed;
 use rstd::prelude::*;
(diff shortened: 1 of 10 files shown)


NicJak:
The individual ProxyType (I would propose e.g. "owner", "employee" with permissions to only edit articles for employees), which decides which functions are callable is still missing at this point.

bazaar/src/lib.rs
```diff
@@ -98,11 +100,11 @@ decl_module! {
             ensure!(<encointer_communities::Module<T>>::community_identifiers().contains(&cid),
                 Error::<T>::InexistentCommunity);
 
-            ensure!(!BusinessRegistry::<T>::contains_key(cid, &sender), Error::<T>::ExistingBusiness);
+            let anonymous_proxy = pallet_proxy::Pallet::<T>::anonymous_account(&sender, &ProxyTypeOf::<T>::default(), 0, None);
```
clangenb, line 103: I am unsure of automatic proxy creation here.

Reasoning Against:
* We enforce the creation of a proxy account, which bloats runtime storage for businesses that don't need that.
* including the proxy pallet, couples the pallets.
* Permission configuration via the proxy filter happens on runtime level. Hence, it does not reduce the work that needs to be done there.

Reasoning Pro:
* It simplifies the client code because. Otherwise, we'd need to wrap a call to the proxy pallet
* if we want to make proxy, we need on extra extrinsic, but which business wants to have a proxy directly upon creation??

For me, the cons over weigh, what do you think @NicJak?

brenzi: I believe proxy creation should be standard. Busnisses may start without, but it is not possible to upgrade your business and make a reverse proxy out of an existing account (is it?)

however: "standard" could also be done by example by our client code. no need for the pallet to do that.

I'm in favor of the pallet not enforcing proxies but the client will default to creating one

brenzi: So, @NicJak may I ask you to remove the proxy creation code?

PR #49, opened 2021-08-09
Fix/bazaar rpc

clangenb:
The bazaar was forgotten to be wired up in the node and was never tested before. Thus, two issues, which are fixed in this PR,  were not found:

* `Copy` trait bound on `AccountId` will never be satisfied with the actual `AccountId` type.
* Hard-coded `DenyUnsafe::Yes` prevented exposure of the new bazaar rpcs even if the flag `rpc-methods unsafe` was set.

Tested: With these patches, I could successfully register a business and retrieve the businesses with the new rpc.


clangenb:
@NicJak FYI, @brenzi 

Merging these tiny details now to proceed with the node.

PR #55, opened 2021-08-20
refactor `new_community` to reduce complexity to O(1) using geohash buckets

### LINKED ISSUE #9: currencies limited to 1000 locations

pifragile:
This PR replaces https://github.com/encointer/pallets/pull/50.
closes #9 

* organize meetup location storage in buckets, addressed by geohash
* reduce minimal distance checks to adjacent buckets

It incorporates all changes discussed in the above PR and adds one more feature:
https://github.com/encointer/pallets/commit/b61b39617e157cfde9026c3bec3e60d225953d77 incorporates the datastructure changes from `communities` (due to geohash) into `ceremonies`. The main problem here was that we can no longer retrieve a simple list of all locations from a community, because there is a layer for geohashes in between. I solved the problem as follows:

1.  Add a function `pub fn get_locations(cid:&CommunityIdentifier) -> Vec<Location>` that uses `iter_prefix_values` on the `Locations` double map in simply concatenates all the vectors corresponding to the geohashes of one community.
2. Because we rely on the indices of the elements in the above retrieved vector in allocation and issuance, I prevented locations from being added/removed in phases other that `CeremonyPhaseType::REGISTERING`.

also, what is crucial for this to work is that `iter_prefix_values` reliably returns its elements in the same order (if the underlying datastructure does not change). Can we rely on that?

note that those changes are somewhat transitional, as the algorithms for issuance and assignment will change again. but i wanted to have a version that works with the current `communities` pallet in order for this PR to be merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1255,8 +1255,9 @@ dependencies = [
 [[package]]
 name = "geohash"
 version = "0.11.1-alpha.0"
-source = "git+https://github.com/encointer/geohash?branch=master#1fd88eb4cd3a2db448d3766677e043800c66385f"
+source = "git+https://github.com/encointer/geohash?branch=master#a4f31108934f0022420c4dbbdc76a3bf60a54a04"
 dependencies = [
+ "parity-scale-codec",
  "substrate-fixed",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1253,21 +1253,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "geo-types"
-version = "0.7.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d8bd2e95dd9f5c8ff74159ed9205ad7fd239a9569173a550863976421b45d2bb"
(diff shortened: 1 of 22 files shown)


brenzi (approved):
LGTM

brenzi:
@pifragile as this is a breaking change, please bump crate versions of all crates in this repo to 0.7.0

PR #61, opened 2021-09-21
upgrade substrate to  558a68f83

brenzi:
* and update to xcm::v1


brenzi:
this one sucks big time:
https://github.com/paritytech/substrate/commit/6be513d663836c5c5b8a436f5712402a1c5365a3

primitives/src/sybil.rs
```diff
@@ -164,10 +164,10 @@ impl PersonhoodUniquenessRating {
 }
 
 pub fn sibling_junction(id: u32) -> (Junction, Junction) {
-    (Junction::Parent, Junction::Parachain(id))
+    (Junction::OnlyChild, Junction::Parachain(id))
```
brenzi, line 167: @clangenb can you please check if this is legit? it makes the build work and the test pass, but not sure about the meaning TBH

brenzi: hmm, actually, test fails

clangenb: hmm, I see that they have changed quite some things, but this seems wrong. I have only checked briefly. I think we should probably use this:

```
MultiLocation {
   parents: 1,
   interior: X1(Parachain(id),
}
```

brenzi: I'm actually not sure we need to switch to v1 at all, but I just couldn't make v0 build anymore

brenzi: works with v1 now. you were spot on

PR #67, opened 2021-10-03
Use Geohash and CRC32 as CommunityIdentifier

NicJak:
I used the size of geohash discussed. Instead of using a hash for the remaining three bytes, I used a CRC. Since only crc16 and crc32 are easily available, I changed the cid size to [u8; 9].


brenzi:
thx. please rebase on master.

9 Bytes is very inefficient for mem allocation, but I think that is nothing to worry about

bazaar/src/tests.rs
```diff
@@ -242,7 +242,7 @@ fn update_inexistent_offering_is_err() {
 
         assert_error(EncointerBazaar::update_offering(Origin::signed(bob()), cid, 1, url1()), Error::<TestRuntime>::InexistentOffering);
         assert_error(EncointerBazaar::update_offering(Origin::signed(alice()), cid, 0, url1()), Error::<TestRuntime>::InexistentOffering);
-        assert_error(EncointerBazaar::update_offering(Origin::signed(alice()), CommunityIdentifier::zero(), 1, url1()), Error::<TestRuntime>::InexistentOffering);
+        assert_error(EncointerBazaar::update_offering(Origin::signed(alice()), [0; 9], 1, url1()), Error::<TestRuntime>::InexistentOffering);
```
brenzi, line 245: may I suggest to use `CommunityIdentifier::default()` instead?

communities/src/tests.rs
```diff
@@ -210,7 +211,7 @@ fn new_community_works() {
             None,
             None
         ));
-        let cid = CommunityIdentifier::from(blake2_256(&(location.clone(), bs.clone()).encode()));
+        let cid = create_community_identifier(location.clone(), &(bs.clone()).encode());
```
brenzi, line 214: It would be more comfortable to just pass `bs` and do the encoding within the fn

and I'd rather have a constructor `CommunityIdentifier::new()` than a helper function

brenzi (changes requested):

NicJak:
Switching to CRC16 would allow us to use [u8, 7] or maybe [u8, 8] with one "reserve" byte... could be used if one day encointer on mars needs different cids.. :D

communities/src/tests.rs
NicJak: The problem I had there is, that bs is a Vec<T::AccountId> and this is not known around the context of communities.rs. Using a completely generic vector parameter inhibits the use of encode(). Maybe this is an issue, that you or @clangenb could quickly fix/explain. :)

clangenb: We can just pass the trait bound `AccountId: Encode`, which will always be fulfilled for the actual `T::AccountId`

communities/src/tests.rs
```diff
@@ -378,8 +313,9 @@ fn remove_location_works() {
         // remove second location
         EncointerCommunities::remove_location(Origin::signed(some_bootstrapper.clone()), cid, location2).ok();
         assert_eq!(EncointerCommunities::locations(&cid, &geo_hash), vec![]);
-        assert_eq!(EncointerCommunities::cids_by_geohash(&geo_hash), vec![]);
-
+        let mut vector = vec![CommunityIdentifier::default()];
```
clangenb, line 316: Is this because the compiler can't infer the type?

if yes, use:
```
let vec: Vec<CommunityIdentifier> = vec![];
```
or
```
let vec = Vec::<CommunityIdentifier>::new();
```

primitives/src/communities.rs
```diff
@@ -30,7 +32,46 @@ pub type CommunityIndexType = u32;
...
+pub struct CommunityIdentifier {
+    location: [u8; 5],
+    bootstrappers: u32,
+}
+
+impl CommunityIdentifier {
+    pub fn new(location: Location, bootstrappers: &[u8]) -> CommunityIdentifier {
+        let geo_hash = GeoHash::try_from_params(location.lat, location.lon, 5).unwrap();
+        let crc_engine = Crc::<u32>::new(&CRC_32_CKSUM);
+        let mut digest = crc_engine.digest();
```
clangenb, line 47: I'd prefer a strongly typed function arg with generic `AccountId`: `bootstrappers: &[AccountId]` and use `encode` internally.

Of course you need to set a suitable trait bound for `AccountId` ;)

communities/rpc/src/lib.rs
```diff
@@ -98,7 +98,7 @@ macro_rules! refresh_cache {
         let mut cid_names: Vec<CidName> = vec![];
 
         cids.iter().for_each(|cid| {
-            $self.get_storage(cid.as_ref()).map_or_else(
+            $self.get_storage(&cid.as_array()).map_or_else(
```
clangenb, line 101: Here you should be able to simply call the borrow `&cid`.

Anyhow, @brenzi can you explain to me, why you used the macro?

Was it to circumvent setting an extensive amount of trait bounds?

clangenb (commented):
Some minor comments

communities/rpc/src/lib.rs
brenzi: With the macro we have less trait bounds on the fundamental Struct so test mocking is easier
tl;dr: I couldn't make it work otherwise

NicJak: couldn't either.. there are ways to just use the memory area of the struct itself with unsafe code, but seeing the size of the method and expressivity, I prefer this solution

primitives/src/communities.rs
```diff
@@ -30,7 +32,46 @@ pub type CommunityIndexType = u32;
 pub type LocationIndexType = u32;
 pub type Degree = I64F64;
 pub type NominalIncome = I64F64;
-pub type CommunityIdentifier = H256;
+
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, Default, RuntimeDebug, PartialOrd, Ord)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct CommunityIdentifier {
+    location: [u8; 5],
+    bootstrappers: u32,
```
brenzi, line 40: I suggest to call this `digest` as it doesn't matter what went into the digest. Maybe we'd like to include more than just bootstrappers in the future

primitives/src/communities.rs
```diff
@@ -30,7 +32,46 @@ pub type CommunityIndexType = u32;
 pub type LocationIndexType = u32;
 pub type Degree = I64F64;
 pub type NominalIncome = I64F64;
-pub type CommunityIdentifier = H256;
+
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, Default, RuntimeDebug, PartialOrd, Ord)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct CommunityIdentifier {
+    location: [u8; 5],
```
brenzi, line 39: its a geohash. I'd call it that

primitives/src/communities.rs
```diff
@@ -30,7 +32,46 @@ pub type CommunityIndexType = u32;
...
+}
+
+impl CommunityIdentifier {
+    pub fn new<AccountId: Encode>(location: Location, bootstrappers: Vec<AccountId>) -> CommunityIdentifier {
+        let geo_hash = GeoHash::try_from_params(location.lat, location.lon, 5).unwrap();
+        let crc_engine = Crc::<u32>::new(&CRC_32_CKSUM);
+        let mut digest = crc_engine.digest();
+        digest.update(&(bootstrappers).encode());
+        CommunityIdentifier {
+            location: [
```
brenzi, line 50: not idiomatic. do:

let mut geohash_cropped = [0u8; 5];
geohash.clone_from_slice(&geohash[0..5]);

primitives/src/communities.rs
brenzi: and rather make it a [u8;4], so you have less trouble downstream

primitives/src/communities.rs
brenzi: or try
let geohash: [u8;5] = Decode::decode(geohash[0..5]);

primitives/src/communities.rs
```diff
@@ -30,7 +32,46 @@ pub type CommunityIndexType = u32;
...
+                geo_hash[0],
+                geo_hash[1],
+                geo_hash[2],
+                geo_hash[3],
+                geo_hash[4],
+            ],
+            bootstrappers: digest.finalize(),
```
brenzi, line 57: here too you could juggle the bytes with encode/decode

primitives/src/communities.rs
```diff
@@ -30,7 +32,46 @@ pub type CommunityIndexType = u32;
...
+    }
+
+    pub fn as_array(self) -> [u8; 9] {
+        [
+            self.location[0],
+            self.location[1],
+            self.location[2],
+            self.location[3],
+            self.location[4],
+            ((self.bootstrappers >> 24) & 0xff) as u8,
```
brenzi, line 68: this doesn't not look very idiomatic. after the above, this will be easy to simplify

brenzi (changes requested):

primitives/src/communities.rs
NicJak: I don't know if the new version is better.. while it does not access the array contents in a hardcoded fashion, it does copy around a lot. I was unable to find a non-copying version of merging two fixed sized arrays of different size that does not require unsafe code

clangenb: Use: https://github.com/canndrew/concat-arrays

communities/src/tests.rs
```diff
@@ -198,71 +198,6 @@ fn new_community_works() {
     });
 }
 
-#[test]
```
brenzi, line 201: why remove this test?

primitives/src/communities.rs
```diff
@@ -50,6 +55,48 @@ pub fn validate_nominal_income(nominal_income: &NominalIncome) -> Result<(), ()>
...
+impl fmt::Debug for CommunityIdentifier {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        fmt(self, f)
+    }
+}
+
+impl CommunityIdentifier {
+    pub fn new<AccountId: Encode>(location: Location, bootstrappers: Vec<AccountId>) -> CommunityIdentifier {
+        let mut geohash_cropped = [0u8; 5];
+        geohash_cropped.clone_from_slice(&GeoHash::try_from_params(location.lat, location.lon, 5).unwrap()[0..5]);
```
clangenb, line 77: Can we be certain that this does not fail? Runtime code must not panic under any circumstances.

communities/src/tests.rs
NicJak: the bucket size was the same resolution as the used geohash.. I do, though not remember whether this was with the final resolution of the geohash. As I understood the test, it verified whether it is possible to register two communities with the same parameters except for location (which deviates slightly). Since the new version would take into account such small differences in location parameter, the test itself would be contradictory to the change.

NicJak:
@clangenb asked for a follow-up commit within the same series in which we constrain the creation of location structs such that no invalid values >90° / <0° are allowed..

primitives/src/communities.rs
```diff
@@ -50,6 +55,48 @@ pub fn validate_nominal_income(nominal_income: &NominalIncome) -> Result<(), ()>
     Ok(())
 }
 
+fn fmt(cid: &CommunityIdentifier, f: &mut Formatter<'_>) -> fmt::Result {
```
brenzi, line 58: I doubt this works as expected because the bytes are already base32 encoded as they are geohash bytes. 
Please write a unit test in this file that verifies that
the first 5 characters for a community with location 0 = lat 48.669, lon -4.329 are `gbsuv`
add other examples in all quadrants with the help of https://www.movable-type.co.uk/scripts/geohash.html

Please always first define the struct and add implementations afterwards (just reverse order of your blocks)

brenzi (changes requested):

NicJak:
> I don't think locations should be checked when creating a cid. If at all, we should verify them in the constructor of a Location struct.

which is what I did

primitives/src/communities.rs
clangenb: When we have resolved potential invalid lon/lat, I would like the unwrap to be an `.expect("Location struct can't contain invalid lon/lat; qeg");`

clangenb:
I have looked at this in greater detail. The relevant code is here:

https://github.com/encointer/pallets/blob/1d31e63cf3c35c0242319a382b4314d4466ffdd7/communities/src/lib.rs#L101-L107

We already to proper verification of everything here. Hence, I think to simplify things, you should just pass the geohash to the `CommunityIdentifier`'s constructor. No need to check anything inside the constructor, and we can spare some computations because we don't calculate the geohash twice.

If we do it as I suggested before, we will overengineer it.

NicJak:
> Ok, so I think we'd probably need to just make the `CommunityIdentifier` constructor fallible. Otherwise, we'll make quite the effort.
> 
> Furthermore, we can never enforce that the `Location` contains valid values only with private fields, as you could always obtain an invalid location with `Location::decode`.

@brenzi would you agree with this version? If yes, I'll implement it

primitives/src/communities.rs
brenzi: we need to remove the unwrap for sure. But I don't get why we need to validate the location before generating the geohash. geohash creation should fail anyway if locations are invalid. I think we do double work here

primitives/src/communities.rs
```diff
@@ -275,4 +333,17 @@ mod tests {
             ))
         );
     }
+
+    #[test]
+    fn create_new_business_is_ok() {
```
brenzi, line 338: strange name ;-) Please correct

primitives/src/communities.rs
NicJak: > we need to remove the unwrap for sure. But I don't get why we need to validate the location before generating the geohash. geohash creation should fail anyway if locations are invalid. I think we do double work here

now you got me confused, as this is - as I understood - what I asked in my last post, whether you are sure you want me to implement this..

NicJak: ah, wou would want to map the result of try_from _params into the communityIdentifierError?

PR #68, opened 2021-10-07
implement new assignment algorithm and lazy issuance

pifragile:


ceremonies/src/lib.rs
```diff
@@ -529,7 +783,7 @@ impl<T: Config> Module<T> {
             let meetup_count = Self::meetup_count((cid, cindex));
             let reward = Self::nominal_income(cid);
 
-            for m in 1..=meetup_count {
+            for m in 0..meetup_count {
```
brenzi, line 786: please respect the rule that indexing begins with 1
This is motivated by the fact that chain state doesn't store zeros, so you cannot know if meetup count is indeed zero or just hasn't been set.
Leave indexing rules as they are

ceremonies/src/lib.rs
```diff
@@ -638,8 +892,8 @@ impl<T: Config> Module<T> {
         meetup_idx: MeetupIndexType,
     ) -> Option<Location> {
         let locations = <encointer_communities::Module<T>>::get_locations(cid);
-        if (meetup_idx > 0) && (meetup_idx <= locations.len() as MeetupIndexType) {
-            Some(locations[(meetup_idx - 1) as usize])
+        if meetup_idx < locations.len() as MeetupIndexType {
```
brenzi, line 895: dito

ceremonies/src/lib.rs
```diff
@@ -653,9 +907,6 @@ impl<T: Config> Module<T> {
         if !(<encointer_scheduler::Module<T>>::current_phase() == CeremonyPhaseType::ATTESTING) {
             return None;
         }
-        if meetup_idx == 0 {
```
brenzi, line 656: dito

ceremonies/src/tests.rs
```diff
@@ -256,12 +255,12 @@ fn perform_bootstrapping_ceremony(
     run_to_next_phase();
     // ATTESTING
     let loc = EncointerCommunities::get_locations(&cid)[0];
-    let time = correct_meetup_time(&cid, 1);
+    let time = correct_meetup_time(&cid, 0);
```
brenzi, line 258: dito....

ceremonies/src/lib.rs
```diff
@@ -82,18 +78,49 @@ decl_storage! {
...
+
+        NewbieRegistry get(fn newbie_registry): double_map hasher(blake2_128_concat) CommunityCeremony, hasher(blake2_128_concat) ParticipantIndexType => T::AccountId;
+        NewbieIndex get(fn newbie_index): double_map hasher(blake2_128_concat) CommunityCeremony, hasher(blake2_128_concat) T::AccountId => ParticipantIndexType;
+        NewbieCount get(fn newbie_count): map hasher(blake2_128_concat) CommunityCeremony => ParticipantIndexType;
+
+
+        AllowedBootstrapperCount get(fn allowed_bootstrapper_count): map hasher(blake2_128_concat) CommunityCeremony => ParticipantIndexType;
+        AllowedReputableCount get(fn allowed_reputable_count): map hasher(blake2_128_concat) CommunityCeremony => ParticipantIndexType;
+        AllowedEndorseeCount get(fn allowed_endorsee_count): map hasher(blake2_128_concat) CommunityCeremony => ParticipantIndexType;
+        AllowedNewbieCount get(fn allowed_newbie_count): map hasher(blake2_128_concat) CommunityCeremony => ParticipantIndexType;
```
brenzi, line 101: Please rename `Allowed` to `Assigned` as this is not a configuration limit but the number of assigned members per actual ceremony

ceremonies/src/lib.rs
```diff
@@ -82,18 +78,49 @@ decl_storage! {
...
+        S1Newbies get(fn s1_newbies): map hasher(blake2_128_concat) CommunityCeremony => u64;
+
+        S2BootstrappersReputables get(fn s2_bootstrappers_reputables): map hasher(blake2_128_concat) CommunityCeremony => u64;
+        S2Endorsees get(fn s2_endorsees): map hasher(blake2_128_concat) CommunityCeremony => u64;
+        S2Newbies get(fn s2_newbies): map hasher(blake2_128_concat) CommunityCeremony => u64;
+
+
+        MBootstrappersReputables get(fn m_bootstrappers_reputables): map hasher(blake2_128_concat) CommunityCeremony => u64;
+        MEndorsees get(fn m_endorsees): map hasher(blake2_128_concat) CommunityCeremony => u64;
+        MNewbies get(fn m_newbies): map hasher(blake2_128_concat) CommunityCeremony => u64;
```
brenzi, line 115: Please introduce a new struct in primitives/ceremonies
```
pub struct AssignmentParams {
  pub m: u64,
  pub s1: u64,
  pub s2: u64
}
```
and derive all usual traits, see other examples

And simplify this storage definition and all function signatures accordingly

brenzi (changes requested):
this is a first stage review. Please fix the mentioned comments before I continue to review
overall, the looks clean. nice work

ceremonies/src/tests.rs
pifragile: done

ceremonies/src/lib.rs
pifragile: done

brenzi:
I'll merge this AFTER we release 0.7.0. Can you please bump the crate versions to 0.8.0 in this PR?

ceremonies/src/lib.rs
```diff
@@ -515,101 +748,95 @@ impl<T: Config> Module<T> {
...
-                        <ParticipantReputation<T>>::insert(
-                            (cid, cindex),
-                            &p,
-                            Reputation::VerifiedUnlinked,
-                        );
-                    }
+
+        let cindex = <encointer_scheduler::Module<T>>::current_ceremony_index() - 1;
+        let reward = Self::nominal_income(cid);
+        let m = Self::get_meetup_index((*cid, cindex), participant)?;
```
NicJak, line 759: that is a very short name for a variable that is used over ten lines.. short variable names are no problem if they're used within very few lines of creation (e.g. a small loop / match statement), the longer a variable is used, the more descriptive should its name be..

ceremonies/src/lib.rs
```diff
@@ -138,14 +157,16 @@ decl_module! {
 
             let cindex = <encointer_scheduler::Module<T>>::current_ceremony_index();
 
-            if <ParticipantIndex<T>>::contains_key((cid, cindex), &sender) {
+            if <BootstrapperIndex<T>>::contains_key((cid, cindex), &sender) ||
+                <ReputableIndex<T>>::contains_key((cid, cindex), &sender)||
+                <EndorseeIndex<T>>::contains_key((cid, cindex), &sender)||
+                <NewbieIndex<T>>::contains_key((cid, cindex), &sender)
```
NicJak, line 163: this checking of presence of the sender in any of the indices probably corresponds to a check of existence or similar.. one could extract this into a method with parameters (cid, cindex, sender) and name the question accordingly..

when changing the data structure (unprobable on blockchain, i know), the phrasing of the method can provide hints of  whether to change the logic in it or not..

ceremonies/src/lib.rs
```diff
@@ -515,101 +748,95 @@ impl<T: Config> Module<T> {
...
+        // first, evaluate votes on how many participants showed up
+        let (n_confirmed, n_honest_participants) =
+            match Self::ballot_meetup_n_votes(cid, cindex, m) {
+                Some(nn) => nn,
+                _ => {
+                    return Err(<Error<T>>::VotesNotDependable.into());
+                }
+            };
+        let meetup_participants = Self::get_meetup_participants((*cid, cindex), m);
+        for p in &meetup_participants {
```
NicJak, line 774: the loop is a little big for just calling it p, I would go for participant

ceremonies/src/lib.rs
```diff
@@ -515,101 +748,95 @@ impl<T: Config> Module<T> {
...
+                if other_participant == p {
+                    continue;
+                }
+                let attestees_from_other = Self::attestation_registry(
+                    (cid, cindex),
+                    &Self::attestation_index((cid, cindex), &other_participant),
+                );
+                if attestees_from_other.contains(&p) {
+                    was_attested_count += 1;
                 }
```
NicJak, line 807: nested loops have high potential of confusion.. the operation you're doing here is a good use case for the iter() api..
pseudo-code: 
let was_attested_count = meetup_participants.iter()
.filter(o -> p!=o)
.map(o -> Self::attestation_index(...))
.map(o -> Self::attestation_registry(...))
.filter(o -> attestees_from_other(p))
.count();

ceremonies/src/lib.rs
```diff
@@ -164,10 +185,38 @@ decl_module! {
                 // register participant as reputable
                 <ParticipantReputation<T>>::insert((cid, cindex),
                     &sender, Reputation::UnverifiedReputable);
+
+                participant_has_reputation = true;
```
NicJak, line 189: wouldn't this correspond to proof.is_some()?

NicJak (commented):
Since you showed interest in the code(-quality) review, I reviewed your code with some remarks. The code is readable and is consistent in its type of writing. (nearly) all variables have meaningful and descriptive names.

ceremonies/src/lib.rs
```diff
@@ -164,10 +185,38 @@ decl_module! {
...
-            <ParticipantIndex<T>>::insert((cid, cindex), &sender, &new_count);
-            <ParticipantCount>::insert((cid, cindex), new_count);
+
+            if <encointer_communities::Module<T>>::bootstrappers(cid)
+                .contains(&sender) {
+                let participant_index = <BootstrapperCount>::get((cid, cindex)).checked_add(1).
+                    ok_or("[EncointerCeremonies]: Overflow adding reputable to registry")?;
+                <BootstrapperRegistry<T>>::insert((cid, cindex), &participant_index, &sender);
+                <BootstrapperIndex<T>>::insert((cid, cindex), &sender, &participant_index);
+                <BootstrapperCount>::insert((cid, cindex), participant_index);
```
NicJak, line 198: these statements are very alike for every if-block.. abstracting them into a function is not meaningful, because the registry, index and count do not have any explicit relation in the code.. would they have been "hierarchical" in a way of having a BootstrapperData-object, a ReputableData-object which each contain the three maps etc., one could have made a function by passing said data object and inserting the arguments accordingly.. (this is more fyi)
in here, I would propose to at least add a comment that states that all if-blocks do basically the same.. this saves time for every developer who has to read the code.. also the entirety of the if-else-if-else-block could be moved to a function...

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+        let num_reputables = Self::reputable_count(community_ceremony);
+        let num_endorsees = Self::endorsee_count(community_ceremony);
+        let num_newbies = Self::newbie_count(community_ceremony);
+
+        if num_locations == 0 {
+            return Err(<Error<T>>::NoLocationsAvailable.into())
+        }
+        let mut num_meetups = min(num_locations, Self::find_prime_below(num_boostrappers + num_reputables));
+
+        let num_assigned_bootstrappers = num_boostrappers;
```
NicJak, line 486: we're in a function with 7 local variables, try not to introduce additional unneccessary variables.. why would you create a unchangeable variable that is the precise same as another?

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
-    // this function is expensive, so it should later be processed off-chain within SubstraTEE-worker
-    // currently the complexity is O(n) where n is the number of registered participants
-    fn assign_meetups() {
+    fn generate_meetup_assignment_params(community_ceremony: CommunityCeremony) -> DispatchResult {
+        let meetup_multiplier = 10u64;
+        let num_locations = <encointer_communities::Module<T>>::get_locations(&community_ceremony.0).len() as u64;
+        let num_boostrappers = Self::bootstrapper_count(community_ceremony);
+        let num_reputables = Self::reputable_count(community_ceremony);
+        let num_endorsees = Self::endorsee_count(community_ceremony);
+        let num_newbies = Self::newbie_count(community_ceremony);
```
NicJak, line 479: moving the creation of a variable towards the first use allows the reader of code to not fill his brain at once.. also it shows more visibly that (e.g. num_locations) certain variables are used only once and can then be "forgotten"

also eliminating local variables is sometimes desirable: maybe num_locations could be eliminated by directly matching the result of get_locations.. only, if the code becomes more readable

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+
+        if num_locations == 0 {
+            return Err(<Error<T>>::NoLocationsAvailable.into())
+        }
+        let mut num_meetups = min(num_locations, Self::find_prime_below(num_boostrappers + num_reputables));
```
NicJak, line 484: I know this will sound contradictory to the two other comments about minimizing the amount of local variables.. but wouldn't a mutable variable that is overwritten exactly once (not iteratively) indicate that these are in fact two different kind of variables and might need different names? isn't this a max_num_meetups?

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+
+        let num_assigned_bootstrappers = num_boostrappers;
+
+        let mut available_slots = num_meetups * meetup_multiplier - num_assigned_bootstrappers;
```
NicJak, line 488: here I think that the mutable variable is correct, as what we basically do is count down on it in a pseudo-loop (first the reputables, then the endorsees etc.)

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+
+        let num_assigned_reputables = min(num_reputables, available_slots);
+        available_slots -= num_assigned_reputables;
+
+        let num_assigned_endorsees = min(num_endorsees, available_slots);
+        available_slots -= num_assigned_endorsees;
+
+        let max_assigned_newbies = (num_assigned_bootstrappers + num_assigned_reputables + num_assigned_endorsees) / 3;
+        let mut num_assigned_newbies = min(num_newbies, available_slots);
+        num_assigned_newbies = min(num_assigned_newbies, max_assigned_newbies);
```
NicJak, line 499: let num_assigned_newbies = min(min(num_newbies, available_slots),max_assigned_newbies);

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+
+        let num_participants = num_assigned_bootstrappers + num_assigned_reputables + num_assigned_endorsees + num_assigned_newbies;
+
+        num_meetups = num_participants / meetup_multiplier;
+
+        // ceil
+        if num_participants % meetup_multiplier != 0 {num_meetups += 1;}
```
NicJak, line 506: use a library ceil function directly on the assignment of value (make sure to use floating point division within the method)

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+
+
+        let assignment_params_bootstrappers_reputables = Self::generate_assignment_function_params(num_assigned_bootstrappers + num_assigned_reputables, num_meetups);
+        let assignment_params_endorsees = Self::generate_assignment_function_params(num_assigned_endorsees, num_meetups);
+        let assignment_params_newbies = Self::generate_assignment_function_params(num_assigned_newbies, num_meetups);
+
```
NicJak, line 512: why introduce variables that are just used once?

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+    }
+
+    fn validate_equal_mapping(num_participants: u64, m: u64, n: u64, s1: u64, s2:u64) -> bool{
+        if num_participants < 2 {
+            return true;
+        }
+
+        let mut meetup_index_count: Vec<u64> = vec![0; n as usize];
+        let mut meetup_index_count_max = (num_participants - m) / n;
+        if (num_participants - m) % n != 0 {meetup_index_count_max += 1;}
```
NicJak, line 594: same here: using a ceil function gives you immutability of a variable.. this is a very good trade.. ;)

ceremonies/src/lib.rs
```diff
@@ -359,148 +414,326 @@ decl_error! {
...
+        loop {
+            s1 = Self::get_random_nonzero_group_element(m);
+            s2 = Self::get_random_nonzero_group_element(m);
+            if Self::validate_equal_mapping(num_participants, m, num_meetups, s1, s2) {
+                break;
+            } else {
+                skip_count += 1;
+                if skip_count > max_skips {
+                    break;
                 }
```
NicJak, line 630: this is a while loop with the condition (skip_count > max_skips) isn't it?

ceremonies/src/lib.rs
pifragile: in order to be consistent with the other `num_assigned_{...}` variables

ceremonies/src/lib.rs
pifragile: ok, which function could i use? and why floating point? i thought everything should be fixed point..

ceremonies/src/lib.rs
pifragile: i think it is more readable like this

ceremonies/src/lib.rs
pifragile: noob question: why is it desirable for a variable to be immutable?

ceremonies/src/lib.rs
pifragile: well, yes with the condition `skip_count <= max_skips` though. but absolutely, idk what i was doing here😂

ceremonies/src/lib.rs
pifragile: agree, not my code. also not really an excuse😅

ceremonies/src/lib.rs
pifragile: agree, not my code. also not really an excuse😅

pifragile:
@NicJak thank you for your review! i added some remarks and questions. no comment means that i agree :) i will leave  your comments unresolved for the moment being, as i will soon have to talk to @brenzi about my future role in the project and then we can take it from there!

ceremonies/src/lib.rs
NicJak: philosophical approach: changing the value of a variable is achieved by either doing some mathematical operation on it or by plainly overwriting it.. in both cases the meaning of the value changes. take the physical world as an example.. let's say we talk about my weight.. 
1. you put me on a scale and read the displayed mass in kg. if you substract the weight of my clothes and store into the same variable, you are inconsistent with what you describe as my weight. is it naked weight or clothed weight. any programmer debugging any function part of code that treats "weight" will have to figure out, which type of weight you meant.. 
2. you put yourself on the same scale and overwrite my weight.. same argument as above.. I will always have to figure out which weight I am working at

I'm sure there are better examples and explanations. also you will find much better literature on it by simply searching for immutability of objets/variables etc online. for example: if you only work with immutable variables, you will be able to parallelize any loop without side-effect (and if side-effects exist, apply this rule recursively, except execution order matters)

brenzi (approved):
subject to approval by @clangenb

ceremonies/src/lib.rs
```diff
@@ -523,28 +523,11 @@ impl<T: Config> Module<T> {
     }
 
     fn find_coprime_below(upper_bound: u64) -> u64 {
```
pifragile, line 525: @NicJak i just stumbled upon this. if you always take `n-1` the algorithm becomes predictable. s1 is supposed to be random and coprime, that's the whole point of the algorithm. but it is my bad, i was not explicit about that when i explained it to you, sorry!

ceremonies/src/lib.rs
```diff
@@ -358,147 +375,465 @@ decl_error! {
...
-			if n.rem_euclid(12) > 0 {
-				n_meetups += 1;
+		for i in 0..max_index {
+			let t1 = (n as i64 * i as i64 + meetup_index as i64 - assignment_params.s2 as i64)
+				.rem_euclid(assignment_params.m as i64);
+			let t2 = Self::mod_inv(assignment_params.s1 as i64, assignment_params.m as i64);
+			let t3 = (t1 * t2).rem_euclid(assignment_params.m as i64);
+			if t3 >= num_participants as i64 {
+				continue
+			}
```
clangenb, line 725: Why all this type casting here? All these functions also exist for `u64`

ceremonies/src/lib.rs
```diff
@@ -629,34 +956,32 @@ impl<T: Config> Module<T> {
...
-		let locations = <encointer_communities::Module<T>>::get_locations(cid);
-		if (meetup_location_idx > 0) &&
-			(meetup_location_idx <= locations.len() as MeetupLocationIndexType)
-		{
-			Some(locations[(meetup_location_idx - 1) as usize])
+		let locations = <encointer_communities::Module<T>>::get_locations(&cc.0);
+		let assignment_params = Self::assignments(cc).locations;
+		let location_idx =
+			Self::assignment_fn(meetup_idx, assignment_params, locations.len() as u64);
+		if (location_idx >= 0) && (location_idx < locations.len() as u64) {
```
clangenb, line 966: `(location_idx >= 0)` is an unnecessary check as this is an unsigned type.

ceremonies/src/lib.rs
```diff
@@ -358,147 +375,465 @@ decl_error! {
...
+	fn get_random_nonzero_group_element(m: u64) -> u64 {
+		let mut random_source =
+			RandomNumberGenerator::<T::Hashing>::new(T::RandomnessSource::random_seed().0);
 
-			// capping the amount a participants prevents assigning more meetups than there are locations.
-			if n > n_locations * 12 {
-				warn!(target: LOG, "Meetup Locations exhausted for cid: {:?}", cid);
-				n = n_locations * 12;
+		// random number in [1, m-1]
+		(random_source.pick_usize((m - 2) as usize) + 1) as u64
```
clangenb, line 676: I think, we can just implement a `pick_u64` on the random number generator to make this more beautiful.

ceremonies/src/tests.rs
```diff
@@ -66,10 +67,13 @@ pub fn set_timestamp(t: u64) {
 }
 
 /// get correct meetup time for a certain cid and meetup
-fn correct_meetup_time(cid: &CommunityIdentifier, mindex: MeetupLocationIndexType) -> Moment {
+fn correct_meetup_time(cid: &CommunityIdentifier, mindex: MeetupIndexType) -> Moment {
 	//assert_eq!(EncointerScheduler::current_phase(), CeremonyPhaseType::ATTESTING);
 	let cindex = EncointerScheduler::current_ceremony_index() as u64;
-	let mlon: f64 = EncointerCeremonies::get_meetup_location(cid, mindex).unwrap().lon.lossy_into();
+	let mlon: f64 = EncointerCeremonies::get_meetup_location((*cid, cindex as u32), mindex)
```
clangenb, line 73: Why transform the cindex to `u64` above to transform it to a `u32` below?

ceremonies/src/tests.rs
```diff
@@ -1015,7 +1081,8 @@ fn get_meetup_time_works(lat_micro: i64, lon_micro: i64) {
 
 		let tol = 60_000; // [ms]
 		assert!(
-			tol > (EncointerCeremonies::get_meetup_time(&cid, 1).unwrap() as i64 - mtime as i64)
+			tol > (EncointerCeremonies::get_meetup_time((cid, cindex), 1).unwrap() as i64 -
+				mtime as i64)
 				.abs() as u64
```
clangenb, line 1086: I think we can remove those type casts to `i64`

ceremonies/src/tests.rs
```diff
@@ -1076,41 +1136,73 @@ fn grow_population_works() {
...
+		// register everybody again. also those who didn't have the chance last time
+		for pair in participants.iter() {
+			let proof = get_proof(cid, cindex - 1, pair);
+			register(account_id(&pair), cid, proof).unwrap();
+		}
+		run_to_next_phase();
 		// ASSIGNING
-		let meetups: Vec<Vec<AccountId>> = get_meetups(cid);
-		assert_eq!(meetups.len(), 2);
-		// whitepaper III-B Rule 3: no more than 1/3 participants without reputation
```
clangenb, line 1104: I would like to keep the whitepaper comment

ceremonies/src/tests.rs
```diff
@@ -1124,126 +1216,377 @@ fn grow_population_works() {
 		run_to_next_phase();
 		// ASSIGNING
 		assert_eq!(proof_count, 13);
+		assert_eq!(EncointerCeremonies::meetup_count((cid, cindex)), 2);
+	});
+}
 
-		let meetups: Vec<Vec<AccountId>> = get_meetups(cid);
-		assert_eq!(meetups.len(), 2);
-		// whitepaper III-B Rule 3: no more than 1/3 participants without reputation
```
clangenb, line 1130: dito

ceremonies/src/tests.rs
```diff
@@ -1124,126 +1216,377 @@ fn grow_population_works() {
...
-		fully_attest_meetups(cid, participants.clone());
+#[test]
+fn find_coprime_below_works() {
+	new_test_ext().execute_with(|| {
+		for i in 2..100 {
+			let coprime = EncointerCeremonies::find_coprime_below(i);
+			assert_eq!(EncointerCeremonies::get_greatest_common_denominator(coprime, i), 1)
+		}
+	});
+}
```
clangenb, line 1244: I like this thorough test!

clangenb: However, I did not see a test for `greatest_common_denominator`?

ceremonies/src/tests.rs
```diff
@@ -1124,126 +1216,377 @@ fn grow_population_works() {
...
+	exp_m_endorsees,
+	exp_m_newbies,
+	exp_n_assigned_bootstrappers,
+	exp_n_assigned_reputables,
+	exp_n_assigned_endorsees,
+	exp_n_assigned_newbies,
+	case(3, 7, 12, 6, 13, 19, 5, 5, 7, 12, 6, 5),
+	case(10, 1, 1, 20, 13, 2, 17, 2, 1, 1, 18, 0)
+)]
+fn generate_meetup_assignment_params_works(
```
clangenb, line 1542: beautiful

clangenb: But maybe we want to put some explicit corner-cases here to be sure that they work?

ceremonies/src/tests.rs
```diff
@@ -1124,126 +1216,377 @@ fn grow_population_works() {
...
-    case(12,48,12*AMOUNT_NEWBIE_TICKETS as usize,0,55, [12; 55].to_vec()),
-)]
-fn assigning_meetup_works(
-	n_bootstrappers: usize,
-	n_reputables: usize,
-	n_endorsees: usize,
-	n_newbies: usize,
-	n_locations: u32,
-	exp_meetups: Vec<usize>,
-) {
```
clangenb, line 1162: Do we not test that anymore?

ceremonies/src/tests.rs
```diff
@@ -1124,126 +1216,377 @@ fn grow_population_works() {
...
+		assert_eq!(EncointerCeremonies::find_prime_below(1), 2);
+		assert_eq!(EncointerCeremonies::find_prime_below(5), 5);
+		assert_eq!(EncointerCeremonies::find_prime_below(10), 7);
+		assert_eq!(EncointerCeremonies::find_prime_below(118), 113);
+		assert_eq!(EncointerCeremonies::find_prime_below(113), 113);
+	});
+}
 
-		// TODO: whitepaper III-B Rule 1: minimize the number of participants that have met at previous ceremony
-		// TODO: whitepaper III-B Rule 2: maximize number of participants per meetup within 3<=N<=12
```
clangenb, line 1142: Track this in #88

clangenb (commented):
Finally, I managed to look at the code. :) It looks very good, but I also have some comments. I reckon this is the essence of my comments.

* I tested that it does indeed not compile with the no due to no_std issues
* Some math stuff I want to see in a dedicated module in `primitives/core`
* Some rust technicalities

General impression:
* it is very technical code, more docs would be really helpful
* I have not double-checked yet, but I think some functions are not unit-tested, and as it's quite complex sometimes. Therefore, I created #89. Additions are welcome, but I don't think we need to do that in this PR. I would like to merge this soon.

ceremonies/src/lib.rs
```diff
@@ -358,147 +375,465 @@ decl_error! {
...
+			s1: Self::find_coprime_below(num_locations),
+			s2: Self::find_prime_below(num_locations),
+		}
+	}
+
+	fn find_coprime_below(upper_bound: u64) -> u64 {
+		let mut range: Vec<u64> = (1..upper_bound).collect();
+		range.shuffle(&mut StdRng::seed_from_u64(Self::get_random_nonzero_group_element(
+			u32::MAX.into(),
+		)));
```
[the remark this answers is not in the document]
clangenb: Why not just use the random permutation algorithm here, as used for the meetup assignments before:

```
    let mut random_source = RandomNumberGenerator::<T::Hashing>::new(
   // we don't need to pass a subject here, as this is only called once in a block.
   // However, without subject this should be initialized outside the community loop, otherwise
   // every community gets the same sequence of permutations if there are the same amount of
   // people per category (bootstrappers, reputables, etc.).
   T::RandomnessSource::random_seed().0,
);

range
   .random_permutation(&mut random_source)
   .unwrap_or_default()
   .into_iter(),
   .find_or_first(...)
...
```

However, this function is deterministic per block, so for the same upper_bound, repetitive calls to this will yield the same result. However, I believe this is not a problem as `find_coprime_below` is only called once per block. So I think a simple comment about that is fine.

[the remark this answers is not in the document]
clangenb: Fix no_std compilation in 3e127900df4603e491910e67c1ccc6ece816b61e

ceremonies/src/lib.rs
```diff
@@ -358,147 +372,424 @@ decl_error! {
...
+		let max_skips = 200;
+		let m = find_prime_below(num_participants);
+		let mut skip_count = 0;
+		let mut s1 = Self::get_random_nonzero_group_element(m);
+		let mut s2 = Self::get_random_nonzero_group_element(m);
+
+		while skip_count <= max_skips {
+			s1 = Self::get_random_nonzero_group_element(m);
+			s2 = Self::get_random_nonzero_group_element(m);
+			if Self::validate_equal_mapping(
```
clangenb, line 636: Randomness here is broken. The randomness source is initialized inside the `get_random_nonzero_group_element` function without. Which, which is always the same for the same block, hence all these subsequent calls `get_random_nonzero_group_element` will yield the same value.

clangenb (approved):
I did some cleanup:

* fixed no_std,
* fixed meetup assignment randomization
* some rust best practices

Follow-ups:
* #90 

Tested that it compiles successfully with the node.

PR #71, opened 2021-10-14
[primitives] fix no_std compilation

clangenb:
@NicJak please always check that your stuff compiles with the encointer-node.


brenzi (approved):
ouch. we need to check this in CI!

PR #73, opened 2021-10-17
[ceremonies/#65] generate random meetup indices

NicJak:
the changes in tests can easily be verified by commenting out the performing the random permutation on the indices (ceremonies/src/lib.rs:487)

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -483,14 +483,15 @@ impl<T: Config> Module<T> {
             }
 
             if !meetups.is_empty() {
+                let mut meetup_indices: Vec<MeetupIndexType>  = (1..=n_locations as MeetupIndexType).collect();
+                meetup_indices =  meetup_indices.random_permutation(&mut random_source).unwrap_or_default();
                 // commit result to state
                 <MeetupCount>::insert((cid, cindex), n_meetups as MeetupIndexType);
                 for (i, m) in meetups.iter().enumerate() {
-                    let _idx = (i + 1) as MeetupIndexType;
                     for p in meetups[i].iter() {
-                        <MeetupIndex<T>>::insert((cid, cindex), p, &_idx);
+                        <MeetupIndex<T>>::insert((cid, cindex), p, &meetup_indices[i]);
                     }
-                    <MeetupRegistry<T>>::insert((cid, cindex), &_idx, m.clone());
+                    <MeetupRegistry<T>>::insert((cid, cindex), &meetup_indices[i], m.clone());
                 }
             };
             debug!(
@@ -526,10 +527,9 @@ impl<T: Config> Module<T> {
         let cids = <encointer_communities::Module<T>>::community_identifiers();
         for cid in cids.iter() {
             let cindex = <encointer_scheduler::Module<T>>::current_ceremony_index() - 1;
-            let meetup_count = Self::meetup_count((cid, cindex));
             let reward = Self::nominal_income(cid);
 
-            for m in 1..=meetup_count {
+            for m in 1..=<encointer_communities::Module<T>>::get_locations(cid).len() as MeetupIndexType {
                 // first, evaluate votes on how many participants showed up
                 let (n_confirmed, n_honest_participants) =
                     match Self::ballot_meetup_n_votes(cid, cindex, m) {
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -92,9 +92,9 @@ decl_storage! {
 
         // all meetups for each ceremony mapping to a vec of participants
         // caution: index starts with 1, not 0! (because null and 0 is the same for state storage)
-        MeetupRegistry get(fn meetup_registry): double_map hasher(blake2_128_concat) CommunityCeremony, hasher(blake2_128_concat) MeetupIndexType => Vec<T::AccountId>;
-        MeetupIndex get(fn meetup_index): double_map hasher(blake2_128_concat) CommunityCeremony, hasher(blake2_128_concat) T::AccountId => MeetupIndexType;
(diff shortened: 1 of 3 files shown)


brenzi (changes requested):
pretty sure this way of doing it won't be compatible with #68. @pifragile can you please comment?

pifragile:
@NicJak @brenzi No this wont be compatible, because the way assignments are handled will be fundamentally changed in https://github.com/encointer/pallets/pull/68. And just out of curiosity, what is the rationale behind randomizing the meetup locations, when we already randomize the pariticipants? Is it about bootstrappers not always being assigned to the first couple locations?

pifragile:
I was just thinking, might there be some nice way to return the locations in a "randomized" order based on some seed input here: https://github.com/encointer/pallets/blob/c2626da67ea0f2dd99be9a636e68f3cf8f1413da/communities/src/lib.rs#L479

brenzi:
not sure if a simple getter should involve random. bad for caching i.e. for geo view explorer

pifragile:
true that makes sense. one could move the functionality to any place actually, was just worried about getting a mess with https://github.com/encointer/pallets/pull/68. maybe we could keep this really simple and for locations not do a full random permutation, but just a  random offset, because unpredictable locations we already get from the assignment process. just thinking aloud :)

pifragile:
i dont understand, a random offset also makes the location unpredictable but takes O(1) and a random permutation takes O(N). i think permutation on locations AND participants is an overkill. what am i missing?

NicJak:
if O(1) is preferred, we could also use an expensive O(1) + C like a hash. One would elaborate the impact of a collision and probably do a small performance comparison as to at what n the break-even-point is..

brenzi:
@pifragile you derive an entire assignment from two primes and a seed, right? why not derive the location permutation the same way?

pifragile:
that would be possible, yes!

ceremonies/src/tests.rs
```diff
@@ -1305,41 +1313,38 @@ fn grow_population_works() {
...
         // ASSIGNING
-        assert_eq!(EncointerCeremonies::meetup_count((cid, cindex)), 2);
-        let meetup3_1 = EncointerCeremonies::meetup_registry((cid, cindex), 1);
-        let meetup3_2 = EncointerCeremonies::meetup_registry((cid, cindex), 2);
+        let meetups: Vec<Vec<AccountId>> = get_meetups(cid);
+        assert_eq!(meetups.len(), 2);
         // whitepaper III-B Rule 3: no more than 1/3 participants without reputation
-        assert_eq!(meetup3_1.len(), 7);
-        assert_eq!(meetup3_2.len(), 6);
+        assert_eq!(get_meetups_by_size(cid, 7).len(), 1);
```
brenzi, line 1342: this a clever trick, but I would prefer something else.
What you actually want to test is the total number of participants assigned. So I'd rather collect all meetups, sum up the number of participants and we're good.

However, #68 will throw this all away so we shouldn't spend too much love for this. Leave as is plz

NicJak: could be easily implemented à la map(meetup -> meetup.len()).sum(), but I'll let it like this, also because this is actually a harder check than just the sum. It also verifies an equal distribution.

NicJak:
how do we proceed with this change?

brenzi:
we should derive a location permutation client side from the randomness we use for assignments. @NicJak If you see how you could do that from the existing code, please go ahead! Otherwise, @pifragile might be able to help

brenzi:
I'll merge this for the 0.6.x releases, before we'll introduce the new assignment logic in 0.7x #68 #75

ceremonies/src/tests.rs
```diff
@@ -313,6 +310,17 @@ fn assert_error(actual: DispatchResult, expected: Error::<TestRuntime>) {
...
+    MeetupRegistry::<TestRuntime>::iter_prefix_values((
+        cid,
+        EncointerScheduler::current_ceremony_index(),
+    )).collect()
+}
+
+fn get_meetups_by_size(cid: CommunityIdentifier, size: usize) -> Vec<Vec<AccountId>> {
+    get_meetups(cid).into_iter().filter(|meetup| meetup.len() == size).collect()
+}
+
```
clangenb, line 323: Just as a general input, collect is relatively expensive, so it might make sense to return iterators here and let the caller decide if we want to collect.

ceremonies/src/lib.rs
```diff
@@ -526,13 +528,13 @@ impl<T: Config> Module<T> {
         let cids = <encointer_communities::Module<T>>::community_identifiers();
         for cid in cids.iter() {
             let cindex = <encointer_scheduler::Module<T>>::current_ceremony_index() - 1;
-            let meetup_count = Self::meetup_count((cid, cindex));
             let reward = Self::nominal_income(cid);
 
-            for m in 1..=meetup_count {
+            for m in MeetupLocationIndex::<T>::iter_prefix_values((cid, cindex))
+                .collect::<BTreeSet<MeetupLocationIndexType>>().iter() {
                 // first, evaluate votes on how many participants showed up
```
clangenb, line 535: Do we need to collect here and turn it into an iterator again? I believe we can do better.

clangenb (commented):
I know this has been merged, but I still wanted to give my two cents here.

In general, this looks good, but I believe, we should be wary about unnecessary `collects` because these are comparatively expensive, and it's a shame if it's done unnecessarily.

ceremonies/src/lib.rs
NicJak: the collection into a BTreeSet removes all duplicates

ceremonies/src/tests.rs
NicJak: Haven't thought of this, but I agree..

PR #92, opened 2021-12-16
[ceremonies/#90] add checked modulo operation

NicJak:


ceremonies/src/lib.rs
```diff
@@ -672,13 +678,19 @@ impl<T: Config> Module<T> {
...
+					.checked_mul(assignment_params.s1)
+					.ok_or(Error::<T>::CheckedMath)?
+					.checked_add(assignment_params.s2)
+					.ok_or(Error::<T>::CheckedMath)?,
+				assignment_params.m,
+			)
+			.ok_or(Error::<T>::CheckedMath)?,
+			n,
+		)
+		.ok_or(Error::<T>::CheckedMath)?;
```
clangenb, line 694: I see you copied not so beautiful code from my side. Allow me to introduce some beatification.

```
		// Simple closure that returns an option to handily use the `?` operator.
		let assign = || -> Option<MeetupIndexType> {
			participant_index
				.checked_mul(assignment_params.s1)?
				.checked_add(assignment_params.s2)
				.map(|div| checked_modulo(div, assignment_params.m))?
				.map(|div| checked_modulo(div, n))?
		};

		assign().ok_or(Error::<T>::CheckedMath)
```

clangenb (commented):
I have a beautification suggestion, it looks good otherwise.

Could you exclude potential overflow/underflow/panics in all other places? If yes, please comment at the relevant places why it is not possible because it is not obvious to me.

ceremonies/src/lib.rs
```diff
@@ -641,24 +649,33 @@ impl<T: Config> Module<T> {
...
 		for i in 0..max_index {
-			let t1 = (n as i64 * i as i64 + meetup_index as i64 - assignment_params.s2 as i64)
-				.rem_euclid(assignment_params.m as i64);
 			let t2 = mod_inv(assignment_params.s1 as i64, assignment_params.m as i64);
-			let t3 = (t1 * t2).rem_euclid(assignment_params.m as i64);
-			if t3 >= num_participants as i64 {
+			let t3 = (n as i64)
+				.checked_mul(i as i64)
+				.and_then(|x| x.checked_add(meetup_index as i64))
+				.and_then(|x| x.checked_sub(assignment_params.s2 as i64))
```
clangenb, line 668: Woah, I kind of forgot about `and_then`. Super useful in this case. 👍

clangenb: You could also use this in the `assignment_fn` and remove my closure instead.

clangenb (commented):
Haha, now you did a very thorough job. 👍 

However, I think in some places the comment is not self-explanatory. What do you think about prefixing them like: `safe; small number per meetup`

clangenb (approved):
Looks good to me, I could follow all the proofs.

PR #98, opened 2021-12-23
parametrize constants

### LINKED ISSUE #95: turn primitives constants into runtime constants
We may want to change these values depending on the chain that carries these pallets, so the constants should be defined in the runtime, not in the pallet:

https://github.com/encointer/pallets/blob/d089a62bc53a39d4fe58af7ae103267f507f4bbb/primitives/src/ceremonies.rs#L199-L200

*Attention*, must replace `2` here with `REPUTATION_LIFETIME + 1` or the reputation will be purged before applied (add a unit test for this!):
https://github.com/encointer/pallets/blob/d089a62bc53a39d4fe58af7ae103267f507f4bbb/ceremonies/src/lib.rs#L968-L969

https://github.com/encointer/pallets/blob/d089a62bc53a39d4fe58af7ae103267f507f4bbb/primitives/src/balances.rs#L39

https://github.com/encointer/pallets/blob/d089a62bc53a39d4fe58af7ae103267f507f4bbb/primitives/src/sybil.rs#L176-L177

https://github.com/encointer/pallets/blob/d089a62bc53a39d4fe58af7ae103267f507f4bbb/primitives/src/communities.rs#L289
*Attention*: unit test that a 10m location distance works

question: Why runtime constants instead of genesis state values?

NicJak:
parametrize numerous constants for pallets as to make them more configurable

closes #95

### DIFF
--- a/balances/src/mock.rs
+++ b/balances/src/mock.rs
@@ -16,9 +16,9 @@
 
 //! Mock runtime for the encointer_balances module
 
-use crate as dut;
-
 use encointer_primitives::balances::Demurrage;
+
+use crate as dut;
 
 use test_utils::*;
 
@@ -26,7 +26,7 @@ type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<TestRunt
 type Block = frame_system::mocking::MockBlock<TestRuntime>;
 
 frame_support::parameter_types! {
-	pub const DefaultDemurrage: i128 =  0x0000000000000000000001E3F0A8A973_i128;
+	pub const DefaultDemurrage: Demurrage = Demurrage::from_bits(0x0000000000000000000001E3F0A8A973_i128);
 }
 
 frame_support::construct_runtime!(
@@ -39,7 +39,7 @@ frame_support::construct_runtime!(
 		Timestamp: pallet_timestamp::{Pallet, Call, Storage, Inherent},
 		EncointerScheduler: encointer_scheduler::{Pallet, Call, Storage, Config<T>, Event},
 		EncointerCommunities: encointer_communities::{Pallet, Call, Storage, Config<T>, Event<T>},
-		EncointerBalances: dut::{Pallet, Call, Storage, Event<T>, Config},
+		EncointerBalances: dut::{Pallet, Call, Storage, Event<T>},
 	}
 );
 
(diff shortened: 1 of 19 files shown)


ceremonies/src/mock.rs
```diff
@@ -47,11 +48,20 @@ frame_support::construct_runtime!(
 	}
 );
 
+parameter_types! {
+	pub const ReputationLifetime: u32 = 1;
+	pub const AmountNewbieTickets: u8 = 50;
+	pub const MinSolarTripTimeS: i32 = 1;
+	pub const DefaultDemurrage: i128 	 = 0x0000000000000000000001E3F0A8A973_i128;
```
clangenb, line 55: fmt: cargo fmt has problems detecting fmt errors in macros because it can't always know if a fmt change, would also change the macros output.

balances/src/mock.rs
```diff
@@ -52,8 +57,10 @@ impl_encointer_scheduler!(TestRuntime);
 // genesis values
 pub fn new_test_ext() -> sp_io::TestExternalities {
 	let mut t = frame_system::GenesisConfig::default().build_storage::<TestRuntime>().unwrap();
-	dut::GenesisConfig { demurrage_per_block_default: Demurrage::from_bits(DEFAULT_DEMURRAGE) }
-		.assimilate_storage(&mut t)
-		.unwrap();
+	dut::GenesisConfig {
+		demurrage_per_block_default: Demurrage::from_bits(DefaultDemurrage::get()),
```
clangenb, line 61: If we have that as a parameter type, we don't need the genesis config value anymore.

clangenb (commented):
Looks good and as expected in general. Only some minor comments.

balances/src/mock.rs
NicJak: removing line 60 results in a failed tests:
tests::transfer_with_demurrage_exceeding_amount_should_fail
tests::demurrage_should_work

clangenb: Sorry for not being clear. Consequently, the genesis value should be removed from the `EncointerBalances` pallet completely and all calls within the pallet that access the genesis field value should fetch the parameter type genesis value.

primitives/src/communities.rs
```diff
@@ -286,7 +286,6 @@ pub mod consts {
 	use fixed::types::{I32F0, U0F64};
 
 	pub const MAX_SPEED_MPS: i32 = 83; // [m/s] max speed over ground of adversary
```
brenzi, line 288: please make this a runtime const too

brenzi (changes requested):
We'll need to temporarly lower adversary speed for close locations for the first phase

brenzi:
plz resolve conflicts so we can merge

brenzi (approved):

PR #100, opened 2021-12-24
Extract assignment logic to separate crate.

clangenb:
I extracted most of the assignment functions that don't need state-access to a separate crate to re-use them in the encointer-client.


brenzi (approved):
I do approve. However, I believe the meetup time calculation is currently redundant an would fit this crate nicely, because the precise meetup time is a direct dependency on the meetup location

clangenb:
Agreed. I extracted the function.

PR #102, opened 2021-12-28
Forbid non-bootsrappers in bootstraping ceremony

### LINKED ISSUE #96: registering endorsees in bootstrapping ceremony should fail
I'm not totally sure, but if I'm not mistaking, it is possible to register endorsees within the first ceremony, which should not be possible?

How to reproduce: in the bot-community.py script, in the init function:
I first initialize bootstrappers (10 bootstrappers)
```    
b = init_bootstrappers(client)
```
then I create a community with the bootstrappers
```
specfile = random_community_spec(b, ipfs_cid)
cid = client.new_community(specfile, b[0])
```
then I endorse a new account:
```
endorse_account(cid,client)
```
then I register all accounts/participants:
```
accounts = client.list_accounts()
register_participants(client,accounts,cid)
```
register_participants works. But shouldn't it handle the case, that endorsees shouldn't be registered on first ceremony?

Checking in the CLI, command: 
```
../target/release/encointer-client-notee list-participants --cid 94QNm8C532sfTnnWgDNT2DrnAZVCZp7u5d9PyAc695Y
```
OUTPUT:
```
listing participants for cid 94QNm8C532sfTnnWgDNT2DrnAZVCZp7u5d9PyAc695Y and ceremony nr 1
number of participants assigned:  11
```
where one out of the 11 participants is the endorsed one.
@brenzi @clangenb

clangenb:
* Closes #96

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -343,6 +343,8 @@ decl_error! {
 	pub enum Error for Module<T: Config> {
 		/// the participant is already registered
 		ParticipantAlreadyRegistered,
+		/// only bootstrappers allowed in first ceremony
+		OnlyBootstrapperAllowed,
 		/// verification of signature of attendee failed
 		BadProofOfAttendanceSignature,
 		/// verification of signature of attendee failed
@@ -408,7 +410,15 @@ impl<T: Config> Module<T> {
 			<BootstrapperRegistry<T>>::insert((cid, cindex), &participant_index, &sender);
 			<BootstrapperIndex<T>>::insert((cid, cindex), &sender, &participant_index);
 			<BootstrapperCount>::insert((cid, cindex), participant_index);
-		} else if is_reputable {
+
+			return Ok(())
+		}
+
+		if cindex == 1 {
+			return Err(Error::<T>::OnlyBootstrapperAllowed)
+		}
+
+		if is_reputable {
 			let participant_index = <ReputableCount>::get((cid, cindex))
 				.checked_add(1)
 				.ok_or(Error::<T>::RegistryOverflow)?;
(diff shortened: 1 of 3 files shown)


ceremonies/src/tests.rs
```diff
@@ -302,7 +302,7 @@ fn assert_error(actual: DispatchResult, expected: Error<TestRuntime>) {
 #[test]
 fn registering_participant_works() {
 	new_test_ext().execute_with(|| {
-		let cid = register_test_community::<TestRuntime>(None, 0.0, 0.0);
+		let cid = perform_bootstrapping_ceremony(None, 1);
```
clangenb, line 305: This test failed after the change because it tries to register participants besides bootstrappers.

ceremonies/src/lib.rs
```diff
@@ -408,7 +410,15 @@ impl<T: Config> Module<T> {
 			<BootstrapperRegistry<T>>::insert((cid, cindex), &participant_index, &sender);
 			<BootstrapperIndex<T>>::insert((cid, cindex), &sender, &participant_index);
 			<BootstrapperCount>::insert((cid, cindex), participant_index);
-		} else if is_reputable {
+
+			return Ok(())
+		}
+
+		if cindex == 1 {
```
brenzi, line 417: this is not legit. non-bootstrappers should only beballowed after the first successful ceremony has beenbperformed by bootstrappers. the cindex will be different for communities joining later

brenzi (changes requested):

ceremonies/src/lib.rs
clangenb: Wow, massive brainfart on my end.

clangenb:
Closing in favor of #134

PR #106, opened 2022-01-05
show pallet parameters as constants in the metadata.

clangenb:
Tested that they appear in the metadata now.


clangenb:
@brenzi, @NicJak FYI

PR #109, opened 2022-01-08
change CommunityMetadata theme type to `IpfsCid`

### LINKED ISSUE #108: refactor community metadata theme outsource to ipfs
A community theme should be a json to make the theme format upgradable without runtime upgrade

just store a field `theme` pointing to a json in ipfs in community pallet metadata storage

theme should be optional

clangenb:
* Closes #108

Tested locally that CI scripts work without any update on the node side.


brenzi (approved):
I wonder if we need icons separately. but it doesn't hurt

clangenb:
I do wonder so, but we could argue that the icons are mandatory while the theme is not.

Issue #84, opened 2021-11-26
Fix assignment when there are not enough locations available


pifragile:
When there are not enough locations available, some participants should not be assigned to a meetup. This can easily be incorporated into `get_meetup_index`.

pifragile:
@brenzi this is the expected behaviour right?

clangenb:
Yes. However, could it be that the current implementation is unfair because the randomized assignment prevents first come-first-served? Or is it even safer to disallow first come-first-served?

Regardless, we recommend anyhow to add more than enough locations when bootstrapping a community. So this theoretical corner-case might not be relevant.

brenzi:
no strict need to be fair as FCFS. just be safe.

NicJak:
The number of assignments (available_slots) is capped by the multiplication of the constant "meetup_multiplier=10u64" with max_num_meetups.. the latter is the minimum number of locations and a prime_below.. thus, as I see it, the number of locations is already limiting the number of assigned participants..

brenzi:
I'm not sure if I get the code right, but while I agree that the number of meetups is capped, this doesn't mean that the participants registrations are cropped to the desired size. Isn't there a wraparound that just causes bigger meetups? Where in the code is the "unregistering" of the excess participants happening?

PR #110, opened 2022-01-09
don't return location index, if not assigned

NicJak:
#84 Fix assignment when there are not enough locations available

since the participant index (a monotonically increasing counter per community_ceremony) is used for filtration, assignement slots are distributed in fifo order


brenzi (approved):
LGTM

clangenb (commented):
LGTM too

PR #111, opened 2022-01-11
[cermonies] fix the purging of the registry

clangenb:
While introducing the parameter types, there was an erroneously introduced parameter.

Randomly found this while chasing a `ProofOutdated` error in the encointer-wallet.


clangenb:
omg, again I was too hasty, some things shouldn't be done too soon before the end of the day... Sorry for that. Should be properly fixed now.

clangenb:
Quite confident that it is correct because of the more thorough testing introduced in the follow-up PR:  https://github.com/encointer/pallets/pull/112/commits/47411c58fe6b9fde7f0d05a7df4d984c3339c858

brenzi (approved):

PR #112, opened 2022-01-11
[Ceremonies] Fix handling of reputation lifetime.

clangenb:
An integer underrun caused an erroneous `ProofOutdated` error when trying to register with a proof of attendance on the app side. (because we increased the reputation lifetime to 337 recently).

Interestingly, the code panics when this happens in the unit tests, whereas nothing was observed when running the node, except for the above `ProofOutdated` error.

**Todo**
* First merge #111


brenzi (approved):
lgtm

Issue #105, opened 2022-01-04
[rpc] account summary queries


brenzi:
In order to reduce communitcation overhead for the mobile app, provide an rpc call which summarizes all information for one account across all communities

* non-zero balances in all currencies (native + communities)
* reputation across all communities (attendance history for the last N ceremonies and bootstrapper status)

might be two calls as two pallets are involved

PR #113, opened 2022-01-15
[ceremonies] add get_reputations rpc

NicJak:
add get_reputations rpc for

#105 [rpc] account summary queries

### DIFF
--- a/ceremonies/rpc/runtime-api/src/lib.rs
+++ b/ceremonies/rpc/runtime-api/src/lib.rs
@@ -1,0 +1,33 @@
+// Copyright (c) 2019 Alain Brenzikofer
+// This file is part of Encointer
+//
+// Encointer is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Encointer is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Runtime API definition required by Ceremonies RPC extensions.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(not(feature = "std"))]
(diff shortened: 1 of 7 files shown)


ceremonies/src/lib.rs
```diff
@@ -397,6 +397,10 @@ decl_error! {
 }
 
 impl<T: Config> Module<T> {
+	fn get_reputations() -> Vec<(CommunityCeremony, T::AccountId, Reputation)> {
```
clangenb: This will need to be public to access it in the node.

clangenb (approved):
I  made `get_repuations` public to access it in the node, but I could then integrate the new rpc in the node: https://github.com/encointer/encointer-node/tree/integrate-reputations-rpc

Compiles, not otherwise tested.

ceremonies/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+use sp_std::vec::Vec;
+
+use encointer_primitives::ceremonies::{CommunityCeremony, Reputation};
+use sp_api::{Decode, Encode};
+
+sp_api::decl_runtime_apis! {
+	pub trait CeremoniesApi<AccountId>
+	where AccountId: Encode + Decode
+	{
+		fn get_reputations() -> Vec<(CommunityCeremony, AccountId, Reputation)>;
```
brenzi, line 31: I thought we want reputation history per account. dumping all accounts doesn't make sense to me

ceremonies/rpc/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+	fn get_reputations(
+		&self,
+		account: AccountId,
+		at: Option<<Block as BlockT>::Hash>,
+	) -> Result<Vec<(CommunityIdentifier, Reputation)>> {
+		self.deny_unsafe.check_if_safe()?;
+
+		let api = self.client.runtime_api();
+		let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
+		return Ok(api
```
clangenb, line 71: @brenzi: Filtering happens on the outside of the runtime. I guess the idea was to filter outside with less limited resources. However, I think iterating twice on the other hand might outweigh the benefit...

brenzi: Doesn't make sense to me why this should be more efficient, unless we use caching. we did use offchain-caching/indexing elsewhere, right?
Caching would be great because reputation only changes once per ceremony

PR #114, opened 2022-01-16
[ceremonies] don't create assignment when less than 3 participan…

### LINKED ISSUE #103: meetup can happen with only one participant.
Probably, this only happens if only one participant is registered for the whole CommunityCeremony.

NicJak:
* Closes #103

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -808,6 +808,9 @@ impl<T: Config> Pallet<T> {
 		let meetup_multiplier = 10u64;
 		let assignment_count =
 			Self::create_assignment_count(community_ceremony, meetup_multiplier)?;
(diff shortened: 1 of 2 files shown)


brenzi (changes requested):
Please add a unit test to see assignment of 2 participants rejected

brenzi (approved):

NicJak:
since you approved the change: have you checked for this fix to be enough or should I proceed with some heuristic testing?

brenzi:
I have approved because I think the testing goes beyond this PR. Would be great if you could add parametric testing and cover a few edge cases of possible small numbers of bootstrappers, endorsees, reputabkes and newbies

clangenb (approved):
LGTM.

Issue #107, opened 2022-01-08
purge inactive communities


brenzi:
maintain a storage that tracks
* how many successul meetups have been performed per cindex

upon new ceremony phase, purge communities which had 0 successful meetups for the last N ceremony phases. N should be a runtime constant

this storage should only be N+1 deep (not store all history)

NicJak:
wouldn't it be easier to implement a number per community that just counts the 0-meetings? This off course wouldn't allow for fancier logic to purge communities, but is far less memory/computation intensive and probable simpler to implement as well..

on_new_phase:
for cid in cids:
  if no_meetups(cid)
     inactivity_counter[cid] ++
     if inactivity_counter[cid]> MAX_INACTIVITY:
       purge(cid)
  else
     inactivity_counter[cid]=0

brenzi:
Yes, that would work. but it won't happen on new phase. Due to lazy issuance, we only know the number of successful meetups at the end of registering / start of assigning. So I guess that is the best moment to check and purge

clangenb:
Closes by #115

PR #115, opened 2022-01-17
Purge inactive communities

NicJak:
* Purge inactive communities as required by #107

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -676,6 +676,11 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn inactivity_counters)]
+	pub(super) type InactivityCounters<T: Config> =
+		StorageMap<_, Blake2_128Concat, CommunityIdentifier, u32>;
+
 	#[pallet::genesis_config]
 	pub struct GenesisConfig<T: Config>
 	where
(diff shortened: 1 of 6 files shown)


ceremonies/src/lib.rs
```diff
@@ -575,6 +582,32 @@ impl<T: Config> Module<T> {
...
+					InactivityCounters::insert(cid, current + 1);
+				}
+			} else {
+				InactivityCounters::insert(cid, 0);
+			}
+		}
+	}
+
+	fn purge_community(cid: CommunityIdentifier) {
+		for cindex in 1..<encointer_scheduler::Module<T>>::current_ceremony_index() {
```
brenzi, line 603: No need to start with 1 here. Too expensive. the registry is purged after reputation lifetime anyway

ceremonies/src/lib.rs
```diff
@@ -575,6 +582,32 @@ impl<T: Config> Module<T> {
...
+			if cindex > T::ReputationLifetime::get() {
```
brenzi, line 604: Db access within a loop. Too expensive

brenzi (changes requested):
Please add unit tests asserting:
* an inactive community gets purged at the expected time
* an active community doesn' t get purged
* purging does remove all meetup locations correctly. And only the right local are purged

ceremonies/src/tests.rs
```diff
@@ -1253,6 +1253,49 @@ fn get_assignment_params_works() {
...
+		let cid1 = CommunityIdentifier::new(
+			Location::new(Degree::from_num(1f64), Degree::from_num(1f64)),
+			Vec::<i64>::new(),
+		)
+		.unwrap();
+
+		let mut cindex = 5;
+		let assignment =
+			AssignmentCount { bootstrappers: 1, reputables: 0, endorsees: 0, newbies: 0 };
+		AssignmentCounts::insert((cid0, cindex), assignment);
```
brenzi, line 1269: #107 demands successful meetups. You are only checking assignment of meetups which might not actually have taken place. I guess you should check Reputation instead oft assignments. Or introduce a flag to signal at least one successful meetup has been verified and issued

brenzi (changes requested):

ceremonies/src/tests.rs
NicJak: how about checking attestationcount!=0 of the last cindex during registration phase?

brenzi: No. Attestations must be validated. Can't be trusted

brenzi: Check Reputation registry or introduce a flag if that is too expensive

brenzi (approved):
LGTM

ceremonies/src/lib.rs
```diff
@@ -766,37 +773,41 @@ impl<T: Config> Pallet<T> {
...
 
-			<IssuedRewards<T>>::remove_prefix((cid, cindex), None);
+		<IssuedRewards<T>>::remove_prefix(cc, None);
+	}
+
+	fn purge_registry(cindex: CeremonyIndexType) {
+		let cids = <encointer_communities::Pallet<T>>::community_identifiers();
+		for cid in cids.iter() {
+			Self::purge_community_ceremony((*cid, cindex));
 		}
```
clangenb, line 811: Even though, copy-dereferencing is already cheap, you can save a bit of efficiency by using `into_iter`, which returns owned values.

clangenb (approved):
Looks good to me!

Thanks for the minor cleanup!

There is one efficiency detail, but I think we can neglect that.

PR #119, opened 2022-01-23
Balances RPC

NicJak:
implement #105 [rpc] account summary queries with complexity O(#communities)

### DIFF
--- a/balances/rpc/src/lib.rs
+++ b/balances/rpc/src/lib.rs
@@ -1,0 +1,91 @@
+// Copyright (c) 2019 Alain Brenzikofer
+// This file is part of Encointer
+//
+// Encointer is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Encointer is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+use jsonrpc_core::{Error, ErrorCode, Result};
+use jsonrpc_derive::rpc;
+use sc_rpc::DenyUnsafe;
+use sp_api::{Decode, Encode, ProvideRuntimeApi};
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{
+	generic::BlockId,
+	traits::{Block as BlockT, Header as HeaderT},
+};
+use std::sync::Arc;
+
+use encointer_balances_rpc_runtime_api::BalancesApi as BalancesRuntimeApi;
+use encointer_primitives::{balances::BalanceEntry, communities::CommunityIdentifier};
+
+#[rpc]
+pub trait BalancesApi<BlockHash, AccountId, BlockNumber>
+where
+	AccountId: 'static + Encode + Decode + Send + Sync,
+	BlockNumber: 'static + Encode + Decode + Send + Sync,
+{
+	#[rpc(name = "encointerBalances_getAllBalances()")]
+	fn get_all_balances(
+		&self,
+		account: AccountId,
+		at: Option<BlockHash>,
+	) -> Result<Vec<(CommunityIdentifier, BalanceEntry<BlockNumber>)>>;
+}
+
+pub struct Balances<Client, Block, AccountId> {
+	client: Arc<Client>,
+	_marker: std::marker::PhantomData<(Block, AccountId)>,
+	deny_unsafe: DenyUnsafe,
+}
+
+impl<Client, Block, AccountId> Balances<Client, Block, AccountId> {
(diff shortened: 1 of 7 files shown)


brenzi (commented):
any idea how we could unit test rpc apis?

balances/src/lib.rs
```diff
@@ -99,6 +99,14 @@ decl_error! {
 }
 
 impl<T: Config> Module<T> {
+	pub fn get_balances(account: &T::AccountId) -> Vec<(CommunityIdentifier, BalanceType)> {
+		let mut balances: Vec<(CommunityIdentifier, BalanceType)> = vec![];
+		for community in <encointer_communities::Module<T>>::community_identifiers().iter() {
+			balances.push((*community, Balance::<T>::get(community, account.clone()).principal));
```
brenzi, line 105: it's not ok to just push the principal
The principal always comes with its individual block height and you need both informations to apply demurrage. As demurrage can differ among communities, I think it would actually be better to apply demurrage directly in this query. It's not such an expensive computation

balances/src/lib.rs
```diff
@@ -99,6 +99,14 @@ decl_error! {
 }
 
 impl<T: Config> Module<T> {
+	pub fn get_balances(account: &T::AccountId) -> Vec<(CommunityIdentifier, BalanceType)> {
```
brenzi, line 102: it's not ok to have `get_balance` and `balance` functions here. both are getters. Please rename to `get_all_balances`

brenzi: or maybe `get_all_balances_demurraged`. see below

brenzi (changes requested):

balances/src/lib.rs
brenzi: and if we return the current balance with demurrage applied, we have to return the block-number for which this applies as well, otherwise, the UI can't update the balance independently over time. As the UI needs to be capable to apply demurrage anyway, we should consider returning values without applying demurrage. 

To summarize the result of my zigzagging: We should return the entire `BalanceEntry` instead of just the principal  `BalanceType` and all is good.

balances/src/lib.rs
```diff
@@ -115,6 +115,15 @@ pub mod pallet {
...
 impl<T: Config> Pallet<T> {
+	pub fn get_all_balances(
+		account: &T::AccountId,
+	) -> Vec<(CommunityIdentifier, BalanceEntry<T::BlockNumber>)> {
+		let mut balances: Vec<(CommunityIdentifier, BalanceEntry<T::BlockNumber>)> = vec![];
+		for community in <encointer_communities::Pallet<T>>::community_identifiers().iter() {
+			balances.push((*community, Balance::<T>::get(community, account.clone())));
+		}
+		return balances
+	}
```
clangenb, line 126: Minor efficiency tweak: you can remove the copy-dereferencing if you use `into_iter`.

balances/src/lib.rs
```diff
@@ -115,6 +115,15 @@ pub mod pallet {
 }
 
 impl<T: Config> Pallet<T> {
+	pub fn get_all_balances(
+		account: &T::AccountId,
+	) -> Vec<(CommunityIdentifier, BalanceEntry<T::BlockNumber>)> {
+		let mut balances: Vec<(CommunityIdentifier, BalanceEntry<T::BlockNumber>)> = vec![];
+		for community in <encointer_communities::Pallet<T>>::community_identifiers().iter() {
+			balances.push((*community, Balance::<T>::get(community, account.clone())));
```
clangenb, line 123: Doesn't this also include entries if the there is no balance in a community? I'd think it is better to only include non-zero entries because most of the users will probably only have community-tokens of few communities.

clangenb (commented):
Code wise, I think it is good. I just think that the general naming is very close to substrate's balances pallet, even if it does not have a custom rpc.

I will think a bit about potential options, but now I don't have a better idea.

balances/rpc/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+use encointer_balances_rpc_runtime_api::BalancesApi as BalancesRuntimeApi;
+use encointer_primitives::{balances::BalanceEntry, communities::CommunityIdentifier};
+
+#[rpc]
+pub trait BalancesApi<BlockHash, AccountId, BlockNumber>
+where
+	AccountId: 'static + Encode + Decode + Send + Sync,
+	BlockNumber: 'static + Encode + Decode + Send + Sync,
+{
+	#[rpc(name = "balances_getBalances")]
```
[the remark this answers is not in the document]
brenzi: That would imply that the balances pallet has these functions

Strictly, it should be 
encointer_balances_getBalance()
Or
encointerBalances_get balance()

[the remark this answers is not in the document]
NicJak: I am not entirely sure, whether I understood the discussion correctly:
I now renamed the string withing the rpc annotation from "balances_getBalances" to "encointerBalances_get_all_balances"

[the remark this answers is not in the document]
brenzi: Sorry, my typo or autocorrect, I think it should be camelcase:
encointerBalances_getAllBalances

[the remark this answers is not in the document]
NicJak: dpme

PR #129, opened 2022-01-25
migrate encointer-ceremonies to frame v2

clangenb:
Metadata has not changed, nothing should break.

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -167,6 +167,8 @@ pub mod pallet {
 			Self::register(cid, cindex, &sender, proof.is_some())?;
 
 			debug!(target: LOG, "registered participant: {:?}", sender);
+
+			Self::deposit_event(Event::ParticipantRegistered(sender));
 			Ok(().into())
 		}
 
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -674,7 +674,8 @@ pub mod pallet {
 
 	#[pallet::genesis_config]
 	pub struct GenesisConfig<T: Config>
-// TODO_MAYBE_WHERE_CLAUSE
+	where
+		<T as pallet_timestamp::Config>::Moment: MaybeSerializeDeserialize,
 	{
 		#[doc = " the default UBI for a ceremony attendee if no community specific value is set."]
 		pub ceremony_reward: BalanceType,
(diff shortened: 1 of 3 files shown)


ceremonies/src/lib.rs
```diff
@@ -56,293 +54,336 @@ use sp_std::{prelude::*, vec};
...
+					Reputation::UnverifiedReputable,
+				);
 			};
 
 			Self::register(cid, cindex, &sender, proof.is_some())?;
 
 			debug!(target: LOG, "registered participant: {:?}", sender);
-			Ok(())
+
+			Self::deposit_event(Event::ParticipantRegistered(sender));
```
clangenb, line 171: With the new pallet syntax it was discovered that we never threw the event! This was probably accidentally deleted sometimes.

ceremonies/src/lib.rs
```diff
@@ -56,293 +54,336 @@ use sp_std::{prelude::*, vec};
...
+			origin: OriginFor<T>,
+			claims: Vec<ClaimOfAttendance<T::Signature, T::AccountId, T::Moment>>,
+		) -> DispatchResultWithPostInfo {
 			let sender = ensure_signed(origin)?;
-			ensure!(<encointer_scheduler::Module<T>>::current_phase() == CeremonyPhaseType::ATTESTING,
-				Error::<T>::AttestationPhaseRequired);
+			ensure!(
+				<encointer_scheduler::Module<T>>::current_phase() == CeremonyPhaseType::ATTESTING,
+				Error::<T>::AttestationPhaseRequired
+			);
```
clangenb, line 184: A lot of fmt because the rustfmt can now analyze the code in the procedural macro. Before, the analyzer did not dare to touch the code, as it might have bricked something.

ceremonies/src/lib.rs
brenzi: Oh. I think we should use this event as it indicates real activity

ceremonies/src/lib.rs
brenzi: Ok

brenzi (approved):

ceremonies/src/lib.rs
clangenb: * I agree, the ceremonies-pallet is very quiet anyhow: #130

Issue #132, opened 2022-01-27
Only use types in pallets implementing `MaxEncodedLen`


clangenb:
With a recent update, substrate changed the default requirements of a pallet's storage and dispatchable arguments. See https://github.com/paritytech/substrate/issues/10652. This default shall become mandatory in the future for all pallets.

For most of our types, we can simply add a derive attribute `MaxEncodedLen`, which specifies the maximum number of bytes that an encoded value can have. To support this in all pallets, we need:

* Derive `MaxEncodedLen` for all types in: 
   * typenum
   * substrate-fixed
   * geohash
* Convert all vecs to either `WeakBoundedVec` or `BoundedVec` and give them a sensible default for the size.
* Change `PalletString` type to a `BoundedVec`, which unfortunately means that we can't use the convenient `String` type in std. 

For the time being, I added the pallet attribute `#[pallet::without_storage_info]` to pallets not yet supporting the `MaxEncodedLen`.

Pallets that still need migration:
- [ ] encointer-bazaar
- [ ] encointer-ceremonies
- [ ] encointer-communities

PR #133, opened 2022-01-28
bump substrate; handle removal of `AccoundId32::default`, add `MaxEncodedLen` support for some pallets

### LINKED ISSUE #120: How to handle removal of `AccountId32::default()`
* https://github.com/paritytech/substrate/pull/10403

Recently, substrate removed the `Default` traitbound for `AccountId32`. This has implications for our storage fields, namely we can't use `T::AccountId` as a value field because the `Default` traitbound is needed (wrapping the field with an `Option<T::AcountId>` results in the same error, which I can't understand why yet).

The `decl_storage` macro does allow setting a default value for types that don't implement a `Default`, but as we are being generic and there is no `const` constructor in the trait bound for `frame_system::Config::Account` we can't set that field to a value without getting an error like: `expected parameter 'AccountId' found AccountId32`.

The only reasonable solution I see to this is to migrate to the new frame v2 pallet syntax. The `Default` traitbound limitation does simply not exist if the storage value query is set to `OptionQuery`.

[Migration Guide](https://crates.parity.io/frame_support/attr.pallet.html#upgrade-guidelines)

EDIT: ~~Storage migration might be needed~~. Seems not to be needed.

### LINKED ISSUE #121: Bump sp-runtime, sp-keyring to v4.1.0-dev
Someone who does not have an up-to-date cargo cache will not be able to build the pallets anymore because substrate has updated the version of `sp-runtime`, `sp-keyring`.

* Blocked by #120

clangenb:
* See #132 
* Closes #120 
* Closes #121

### DIFF
--- a/bazaar/src/lib.rs
+++ b/bazaar/src/lib.rs
@@ -48,6 +48,7 @@ pub mod pallet {
 
 	#[pallet::pallet]
 	#[pallet::generate_store(pub (super) trait Store)]
+	#[pallet::without_storage_info]
 	pub struct Pallet<T>(PhantomData<T>);
 
 	#[pallet::call]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1507,7 +1507,8 @@ dependencies = [
 
 [[package]]
 name = "geohash"
-version = "0.11.1-alpha.0"
+version = "0.12.0"
+source = "git+https://github.com/encointer/geohash?tag=v0.12.0#eed7dcb86ff73ee2689f5f4e091feb86078c11dd"
 dependencies = [
  "parity-scale-codec",
  "scale-info",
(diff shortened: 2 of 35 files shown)


bazaar/src/lib.rs
```diff
@@ -48,6 +48,7 @@ pub mod pallet {
 
 	#[pallet::pallet]
 	#[pallet::generate_store(pub (super) trait Store)]
+	#[pallet::without_storage_info]
 	pub struct Pallet<T>(PhantomData<T>);
```
clangenb, line 52: Substrate will make this forbidden in the future. By then, we need to have completely solved #132.

brenzi (approved):
LGTM

Some things are nicer now, others seem to get more complicated.....

PR #134, opened 2022-02-01
allow bootstrappers only after inactive cycle

### LINKED ISSUE #96: registering endorsees in bootstrapping ceremony should fail

NicJak:
once a community has had a cycle without any successful meetup (measured by issuedrewards), only bootstrappers are allowed for to register for meetups. same is true for new communities
resolves #96 registering non-bootstrappers in bootstrapping ceremony should fail


brenzi:
Not sure this is desired. If a community has 1Mio participants and for some reason one entire ceremony can't take place (24h power outage, all phones off). It would be undesirable to restart the community at 10 participants. Moreover, if reputation lifetime is >2 the reputation would survive to the 2nd ceremony after the interruption and all 1Mio are back in the game.

I basically agree with the resetting logic, but the reputation lifetime should affect the reset holdoff

ceremonies/src/lib.rs
```diff
@@ -739,6 +741,8 @@ impl<T: Config> Pallet<T> {
 			<BootstrapperRegistry<T>>::insert((cid, cindex), &participant_index, &sender);
 			<BootstrapperIndex<T>>::insert((cid, cindex), &sender, &participant_index);
 			<BootstrapperCount<T>>::insert((cid, cindex), participant_index);
+		} else if !Self::rewards_issued((cid, cindex - 1)) {
```
brenzi, line 744: This check should ratger be: no issuance for the last N ceremonies, where N = ReputationLifetime

brenzi (changes requested):

ceremonies/src/lib.rs
NicJak: the described functionality is implemented by inactivity_counters. But then we have the issue, that new communities either accept all registrations initially or (if the counter is initialized at reputation lifetime), that new communities will be purged withing "inactivity_timeout - reputation_lifetime"

brenzi: true: inactivity leads to purging of the community, so we're back at bootstrappers only anyway.
But here's a suggestion how to achieve my requirement:

If the first ceremony has taken place, total_issuance will be > 0 and will always be until an eventual purge.

So why don't you just check this?
if !(total_issuance(cid) > 0): only bootstrappers

NicJak: Wouldn't the total_issuance be a dependency on the balances pallet? If so, that would give us circular dependency..

brenzi: Ceremonies depends on balances anyway. But not the other way round. i don't see the problem

ceremonies/src/lib.rs
```diff
@@ -739,6 +741,8 @@ impl<T: Config> Pallet<T> {
 			<BootstrapperRegistry<T>>::insert((cid, cindex), &participant_index, &sender);
 			<BootstrapperIndex<T>>::insert((cid, cindex), &sender, &participant_index);
 			<BootstrapperCount<T>>::insert((cid, cindex), participant_index);
+		} else if !(<encointer_balances::Pallet<T>>::total_issuance(cid) > 0) {
+			return Err(Error::<T>::OnlyBootstrappers)
```
clangenb, line 745: Elegant solution!

ceremonies/src/lib.rs
NicJak: I did now implement as described, though I must pronounce that I find the dependency of ceremonies on balances quite bad. I didn't see that the dependency existed beforehand, because to me it was clear that there shouldn't be one. In my opinion, it is not the task of ceremonies to imperatively issue rewards after ceremony attendation. Rather ceremonies should offer a service that informs you on the successful attendation of a member. A different component - lets call it free_income component should observe on attendands and grant them the issued rewards.
On a more meta level: it's the communities task to know its members, it's during socializing that you decide who is part of your community, it's the bank/accountants job to manage your equity..

brenzi: yes, separation of concerns would be cleaner as you propose. But for this PR it is out of scope to fix that

brenzi (approved):
LGTM

PR #136, opened 2022-02-10
Council-permissioned communities and ceremony management

### LINKED ISSUE #77: council-permissioned community and ceremony management
currently, everyone can register a new community and other permissioned functions are not really in line with the new common good governance through a council origin

https://github.com/paritytech/polkadot/blob/7a532985a3ccdf4f0c2fbc4b50e1e0806937391f/runtime/polkadot/src/lib.rs#L183-L187

An example how to use this generically in a pallet:
https://github.com/paritytech/substrate/blob/1b646b219e0e0c7489a2a8488381fc2e32da6a1c/frame/identity/src/lib.rs#L144
https://github.com/paritytech/substrate/blob/1b646b219e0e0c7489a2a8488381fc2e32da6a1c/frame/identity/src/lib.rs#L828

The following dispatchables should only be executed upon council motion, so we need to introduce a special origin that will be configured in the runtime instanciation

# communities
all of these should require council origin and be feeless

new_community, currently unpermissioned
https://github.com/encointer/pallets/blob/47282e2a5fe856884fcc875c89a8bdc8fc317435/communities/src/lib.rs#L82-L90

and bootstrappers can add/remove new locations:
https://github.com/encointer/pallets/blob/47282e2a5fe856884fcc875c89a8bdc8fc317435/communities/src/lib.rs#L139-L144

and it takes root origin to migrate metadata, demurrage, UBI:
https://github.com/encointer/pallets/blob/47282e2a5fe856884fcc875c89a8bdc8fc317435/communities/src/lib.rs#L203-L204

# scheduler
all of these should require council origin and be feeless

It currently takes a genesis-defined CeremonyMaster to:

modify ceremony schedule

https://github.com/encointer/pallets/blob/47282e2a5fe856884fcc875c89a8bdc8fc317435/scheduler/src/lib.rs#L80-L82

https://github.com/encointer/pallets/blob/47282e2a5fe856884fcc875c89a8bdc8fc317435/scheduler/src/lib.rs#L89-L91

clangenb:
* Closes #77 

Changes:
* [communities]:
   * Replace `community_master` genesis field with `CommunityMaster` associated type
   * Some extrinsics require now `CommunityMaster` origin, which were unpermissioned before
* [scheduler]
   * Replace `ceremony_master` genesis field with `CeremonyMaster` associated type
* [ceremonies]
   * Remove `ceremonies::grant_reputation` extrinsic

**Note:** has not been tested with the node, but it would require some version bumping there, which would also require a further version bump in the api-client. (yes, again). So I suggest we merge that untested.

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -90,28 +90,6 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		#[pallet::weight(10_000)]
-		pub fn grant_reputation(
-			origin: OriginFor<T>,
-			cid: CommunityIdentifier,
-			reputable: T::AccountId,
-		) -> DispatchResultWithPostInfo {
-			let sender = ensure_signed(origin)?;
-
-			let master = <encointer_scheduler::Pallet<T>>::ceremony_master()
-				.ok_or(Error::<T>::AuthorizationRequired)?;
-			ensure!(sender == master, Error::<T>::AuthorizationRequired);
-
-			let cindex = <encointer_scheduler::Pallet<T>>::current_ceremony_index();
-			<ParticipantReputation<T>>::insert(
-				&(cid, cindex - 1),
-				reputable,
-				Reputation::VerifiedUnlinked,
-			); //safe; cindex comes from within, will not overflow at +1/d
-			info!(target: LOG, "granting reputation to {:?}", sender);
-			Ok(().into())
-		}
-
-		#[pallet::weight(10_000)]
 		pub fn register_participant(
 			origin: OriginFor<T>,
 			cid: CommunityIdentifier,
--- a/communities/src/lib.rs
+++ b/communities/src/lib.rs
@@ -343,35 +343,6 @@ pub mod pallet {
 	#[pallet::getter(fn nominal_income)]
 	pub type NominalIncome<T: Config> =
 		StorageMap<_, Blake2_128Concat, CommunityIdentifier, NominalIncomeType, ValueQuery>;
-
-	#[pallet::storage]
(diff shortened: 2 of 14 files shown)


ceremonies/src/lib.rs
```diff
@@ -89,28 +89,6 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
-		#[pallet::weight(10_000)]
-		pub fn grant_reputation(
-			origin: OriginFor<T>,
-			cid: CommunityIdentifier,
-			reputable: T::AccountId,
-		) -> DispatchResultWithPostInfo {
```
clangenb, line 97: I did not see any reason to keep this, do you @brenzi?

brenzi: No

communities/src/lib.rs
```diff
@@ -60,6 +60,10 @@ pub mod pallet {
 	#[pallet::config]
 	pub trait Config: frame_system::Config + encointer_scheduler::Config {
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// Required origin for adding or updating a community (though can always be Root).
+		type CommunityMaster: EnsureOrigin<Self::Origin>;
```
brenzi, line 65: yeah, love substrate!

brenzi (changes requested):
Looks really nice and lean!

It is a lot of responsibiltiy we give to bootstrappers, allowing them to add and remove locations. I think we will need a community governance to make this a group decision . 
Issue to solve later: #137 

I suggest to play safe and request CeremonyMaster for adding and removing locations as well. Just like metadata and demurrage rate. We can leave that to communities once we have governance #137

PR #148, opened 2022-02-16
better demurrage documentation

brenzi:


clangenb (approved):
I guess the prose can still be improved, but it helps the understanding much better now. :)

PR #150, opened 2022-02-24
add more events & details

### LINKED ISSUE #130: add more events to ceremonies pallet
The ceremonies has only one event currently: `ParticipantRegistered`. We can for sure enrich the events there, e.g.:

* `EncorsedNewcomer`
* diversify `ParticipantRegistered` into `BootstrapperRegistered`, `ReputableRegistered`, `EndorseeRegistered`, `NewbieRegistered`
* `RewardsClaimed`

And some others.

brenzi:
closes #130 

* added events and respective tests to assert their deposited
* improved logging

### DIFF
--- a/ceremonies/src/tests.rs
+++ b/ceremonies/src/tests.rs
@@ -36,7 +36,7 @@ use sp_core::{sr25519, Pair, H256, U256};
 use sp_runtime::traits::BlakeTwo256;
 use std::ops::Rem;
 use test_utils::{
-	helpers::{account_id, bootstrappers, register_test_community, last_event},
+	helpers::{account_id, bootstrappers, last_event, register_test_community},
 	*,
 };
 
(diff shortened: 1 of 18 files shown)


test-utils/src/helpers.rs
```diff
@@ -73,14 +73,13 @@ where
 	CommunityIdentifier::new(location, bs).unwrap()
 }
 
-pub fn assert_last_event<T: frame_system::Config>(
-	generic_event: <T as frame_system::Config>::Event,
-) {
+pub fn last_event<T: frame_system::Config>() -> Option<T::Event> {
```
brenzi, line 76: I believe assertions should only be made within the test routine, otherwise it will not be clear what caused the assertion to  fail

ceremonies/src/tests.rs
```diff
@@ -301,6 +301,9 @@ fn registering_participant_works() {
 		assert_eq!(EncointerCeremonies::bootstrapper_count((cid, cindex)), 0);
 		assert_ok!(register(alice.clone(), cid, None));
 
+		assert_eq!(last_event::<TestRuntime>(),
```
brenzi, line 304: @clangenb this assertion fails because no event was dispatched. I don't understand why that is the case (the real chain dispatches this event, I checked)

test-utils/src/helpers.rs
clangenb: I tend to disagree;

* When it fails, one needs to investigate properly anyhow. So I prefer readability over the exact line of code where it fails.
* Also, upstream changes need to be incorporated, which means now we might have to update every event assertion whereas otherwise adjusting the function might have sufficed.

However, no strong opinion here...

brenzi: The priblem is that you don't get a trace. If you call that fn twice in one test, you don't easily find out if the error occured the first or the second time

Issue #56, opened 2021-08-25
benchmarking for all pallets


brenzi:
we have benchmarking for `communities` but not yet for others. We'll need to set all weights so we need benchamrking

clangenb:
@pifragile these are the leftover tasks for the communities and scheduler pallet: #174

pifragile:
> @pifragile these are the leftover tasks for the communities and scheduler pallet: #174

👍🏽

PR #154, opened 2022-03-01
Benchmarking WIP for communities and scheduler pallet.

NicJak:
Add benchmarking according to #56 benchmarking for all pallets


clangenb (approved):
It looks like a good WIP.

Will merge this as is, but there will be some follow-up tasks: #174 

For now, I disabled the execution of the communities-benchmark with cargo test. It takes an hour to execute.

PR #160, opened 2022-03-03
[communities/#156] rename icons to assets

### LINKED ISSUE #156: rename icons to assets
https://github.com/encointer/pallets/blob/f7d40c9fac59eea8042a465f80221176160c6625/primitives/src/communities.rs#L196

As we'd like to use sounds too, icons is not correct. But it should still be one ipfs folder with all the assets included with standardized filenames/folders

https://github.com/encointer/encointer-wallet-flutter/issues/444

pifragile:
* Closes #156

### DIFF
--- a/primitives/src/communities.rs
+++ b/primitives/src/communities.rs
@@ -193,7 +193,7 @@ pub struct CommunityMetadata {
 	/// utf8 encoded abbreviation of the name
 	pub symbol: PalletString,
 	/// ipfs cid for multi-resolution resource for the community icon
-	pub icons: IpfsCid,
+	pub assets: IpfsCid,
 	/// ipfs cid for style resources
 	pub theme: Option<IpfsCid>,
 	/// optional link to a community site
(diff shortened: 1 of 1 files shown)


primitives/src/communities.rs
```diff
@@ -193,7 +193,7 @@ pub struct CommunityMetadata {
 	/// utf8 encoded abbreviation of the name
 	pub symbol: PalletString,
 	/// ipfs cid for multi-resolution resource for the community icon
```
clangenb, line 195: Please also update the doc string.

clangenb (commented):
Looks goot, but one minor detail.

primitives/src/communities.rs
clangenb: I should be something like:

`IPFS cid to assets necessary for community branding.`

clangenb (approved):
Noice, looks good now!

PR #161, opened 2022-03-04
balances/communities dependency inversion

pifragile:
When implementing the purging of balances in the process of purging a community, I got a cyclic dependency between the communities and balances pallet. Previously, the balances pallet depended on the communities pallet and now I needed the communities pallet to also depend on the balances pallet.

In consultation with @clangenb, we decided that from a domain logic point of view it makes more sense for the communities pallet to depend on the balances pallet.

For this change it was necessary to move `DemurragePerBlock` to the balances pallet and `get_all_balances` to the communities pallet (along with the corresponding RPC endpoint.)


balances/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "pallet-encointer-balances"
 version = "0.8.1"
 authors = ["encointer.org <alain@encointer.org> and Parity Technologies <admin@parity.io>"]
-edition = "2018"
+edition = "2021"
```
clangenb, line 5: Nice 👍

balances/src/tests.rs
```diff
@@ -24,10 +24,9 @@ use encointer_primitives::{
 };
 use frame_support::{assert_noop, assert_ok, traits::OnInitialize};
 use mock::{new_test_ext, EncointerBalances, System, TestRuntime};
-use test_utils::{
-	helpers::{last_event, register_test_community},
-	AccountKeyring,
-};
+use test_utils::{helpers::last_event, AccountKeyring};
+
+use sp_std::str::FromStr;
```
clangenb, line 29: We want to prevent empty lines in the import sections because this screws up the formatter.

communities/rpc/runtime-api/src/lib.rs
```diff
@@ -22,15 +22,22 @@
 use sp_std::vec::Vec;
 
 use encointer_primitives::{
+	balances::BalanceEntry,
 	common::PalletString,
 	communities::{CommunityIdentifier, Location},
 };
 
+use sp_api::{Decode, Encode};
```
clangenb, line 30: Again, we like to keep all imports in one group.

communities/rpc/src/lib.rs
```diff
@@ -44,26 +50,40 @@ pub trait CommunitiesApi<BlockHash> {
 		cid: CommunityIdentifier,
 		at: Option<BlockHash>,
 	) -> Result<Vec<Location>>;
+
+	#[rpc(name = "communities_getAllBalances")]
+	fn communities_get_all_balances(
```
clangenb, line 55: Haha @brenzi, we could get rid of the irregularities here. ;)

communities/src/tests.rs
```diff
@@ -26,6 +28,8 @@ use test_utils::{
 	*,
 };
 
+use encointer_primitives::balances::BalanceType;
```
clangenb, line 31: I guess this is the IDE's automatic import, which does not work as we want it to. :)

communities/src/tests.rs
```diff
@@ -890,3 +897,39 @@ fn get_locations_works() {
 		assert_eq!(result, expected_result);
 	});
 }
+
+#[test]
+fn get_all_balances_works() {
```
clangenb, line 902: Thanks for the new test. 🦖

clangenb (commented):
Looks very good, only some minor comments.

And I will only approve if the integration into the node was successful.

communities/src/tests.rs
pifragile: ✌🏽

communities/src/tests.rs
pifragile: no i think i made this import manually. what's wrong with it?

clangenb: That it enters the newline, and it does not group it with the rest

pifragile: oh i see :)

communities/rpc/src/lib.rs
brenzi: yeah, let's prefix all our rpcs with `encointer`, not with the pallet name

brenzi: so, `encointer_getAllBalances`

brenzi:
To challenge this decision: The `ceremonies` pallet depends on all others anyway. Why not move these functions and storages over there? Not sure what is better, so if you agree on the proposed approach I won't insist

pifragile:
> To challenge this decision: The `ceremonies` pallet depends on all others anyway. Why not move these functions and storages over there? Not sure what is better, so if you agree on the proposed approach I won't insist

I see your point, we would have one dependency less with this approach. For me (with my very limited experience with the project) it intuitively makes sense like we did it now. I would argue that all balance logic that is connected to the balances of specific users or communities (like purging balances) should be done from the communities pallet, because ceremonies don't have anything to do with this process (as opposed to issuance for example). But I am by no means an expert. What do you think @clangenb ?

clangenb:
I think the current implementation is better. The ceremonies-pallet should not become more complex. IMO it handles to much already. I would even go one step further; Remove the dependency on the balances-pallet in the ceremonies pallet and proxy the `issue_rewards` call via the communities-pallet.

communities/rpc/src/lib.rs
clangenb: I agree, with `encointer_getAllBalances`

brenzi:
We could think about abstracting all dependencies using callbacks configurable in the runtime, like `OnCommunityPurged` or `OnPersonhoodVerified`

clangenb (approved):
Looks good to me:

Will merge this with the leftover task:

* #163

PR #162, opened 2022-03-04
Align crate versions

clangenb:
I got some issues while patching the crates in the encointer node because some versions were bumped, while others weren't.


brenzi:
We should define a policy here. Do we really want to bump crate versionjust because some other crate in the same repo had changes?

clangenb:
I don't really know. The trouble did not end there anyhow: https://github.com/scs/substrate-api-client/issues/218

We should finally define a systematic methodology on dependency pinning: branch, tag.

Our setup is too brittle, I can't test #161 in the node now because I can't patch due to version inconsistencies.

I guess I am ready for the overhead of maintaining multiple branches if we can ensure that builds are not suddenly broken.

clangenb:
I will close dangling pr,

PR #164, opened 2022-03-07
[communities/#152] allow communityMaster origin to purge a community

### LINKED ISSUE #152: allow CommunityMaster origin to purge a community
motivation: https://github.com/encointer/encointer-wallet-flutter/issues/410

but anyway, we may want to have that option in the beginning

add a dispatchable `purge_community(cid)` which requires `CeremonyMaster` origin
* purge all locations
* purge all balances

check the inactivity timeout which already does the purging too (but doesn't yet purge balances as it seems - which it should. fix that in the process):
https://github.com/encointer/pallets/blob/4fac5e342055cfb30a6f9eb91f3eec3d03316f35/ceremonies/src/lib.rs#L954

pifragile:
* Closes #152


balances/src/lib.rs
```diff
@@ -243,6 +243,11 @@ impl<T: Config> Pallet<T> {
 	pub fn set_demurrage(cid: &CommunityIdentifier, demurrage: Demurrage) {
 		<DemurragePerBlock<T>>::insert(cid, &demurrage);
 	}
+
+	pub fn purge_balances(cid: CommunityIdentifier) -> DispatchResult {
+		<Balance<T>>::remove_prefix(cid, None);
+		Ok(())
+	}
```
clangenb, line 250: Why the return type if this cannot fail?

communities/src/lib.rs
```diff
@@ -456,6 +466,8 @@ impl<T: Config> Pallet<T> {
 		<CommunityMetadata<T>>::remove(cid);
 
 		<NominalIncome<T>>::remove(cid);
+
+		<encointer_balances::Pallet<T>>::purge_balances(cid).ok();
 	}
```
clangenb, line 471: Haha, I guess this is a hack to remove the unused result warning? :)

Just change the return type as advised above, and then you can remove that.

communities/src/tests.rs
```diff
@@ -450,9 +452,16 @@ fn remove_community_works() {
...
+		assert_ok!(EncointerBalances::issue(cid, &alice, BalanceType::from_num(100)));
+		assert_ok!(EncointerBalances::issue(cid2, &alice, BalanceType::from_num(100)));
+		assert_eq!(EncointerCommunities::get_all_balances(&alice).len(), 2);
+
 		// remove first location
 		EncointerCommunities::remove_community(cid);
 
+		// assert that balances have been purged
+		assert_eq!(EncointerCommunities::get_all_balances(&alice).len(), 1);
+
```
clangenb, line 464: Nice, that you thought of this test!

clangenb (commented):
Minor remark, looks good otherwise!

balances/src/lib.rs
pifragile: haha just copy pasted from another function tbh :) makes sense!

communities/src/tests.rs
pifragile: ✌🏽

brenzi (approved):

clangenb (approved):
Looks good to me now!

PR #165, opened 2022-03-07
[ceremonies/#149] meetup time offset

### LINKED ISSUE #149: implement signed meetup time offset to tune or vary the exact daytime relative to high sun
Story:
* Initially, we'd like the meetup in Zurich to happen at 12:50, not at high sun
* Over time, we may want to shift the meetup time to maximize convenience for all active encointer communities

optional extensions for later (just keep this in mind when designing the solution here): 
* allow drifting: slightly change meetup daytime from ceremony to ceremony within a bounded range. This could be more inclusive overall because everyone will get a chance who has strong time constraints (could be achieved ba manually change this offset after each ceremony)
* allow warping: meetup time moves faster or slower than the sun. This would allow tuning the dateline double-attendance risk

side effects: will require changes in
* encointer-wallet-flutter
* encointer-js
* encointer-node (client)
* encointer-parachain

pifragile:
* Closes #149

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -418,6 +418,9 @@ pub mod pallet {
 			meetup_time_offset: T::Moment,
 		) -> DispatchResultWithPostInfo {
 			<T as pallet::Config>::CeremonyMaster::ensure_origin(origin)?;
+			if <encointer_scheduler::Pallet<T>>::current_phase() != CeremonyPhaseType::REGISTERING {
+				return Err(<Error<T>>::WrongPhaseForChangingMeetupTimeOffset.into())
+			}
 			<MeetupTimeOffset<T>>::put(meetup_time_offset);
 			Ok(().into())
 		}
@@ -490,6 +493,8 @@ pub mod pallet {
 		CheckedMath,
 		/// Only Bootstrappers are allowed to be registered at this time
 		OnlyBootstrappers,
+		/// MeetupTimeOffset can only be changed during registering
+		WrongPhaseForChangingMeetupTimeOffset,
 	}
 
 	#[pallet::storage]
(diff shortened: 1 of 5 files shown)


ceremonies/src/tests.rs
```diff
@@ -1123,6 +1124,17 @@ fn get_meetup_time_works(lat_micro: i64, lon_micro: i64) {
...
+			Origin::signed(master()),
+			Moment::from(100_000u64),
+		)
+		.ok();
+		assert!(
+			tol > (EncointerCeremonies::get_meetup_time(location).unwrap() as i64 -
+				(mtime + 100_000u64) as i64)
+				.abs() as u64
+		);
 	});
```
clangenb, line 1138: Here, I wonder if we should do a separate test for that feature, but this would be quite some duplicate code currently for the test-setup... may be just leave it at that.

clangenb (approved):
Very swift implementation.

brenzi (approved):
lgtm. We shall integrate this in encointer-node#184

brenzi:
oops, wait! changing the meetup time during ASSIGNING or ATTESTING will have strange consequences. please enforce that a change is  only possible during REGISTERING

brenzi (changes requested):
please enforce that a change is only possible during REGISTERING

ceremonies/src/lib.rs
```diff
@@ -411,6 +411,16 @@ pub mod pallet {
...
 			Ok(().into())
 		}
+
+		#[pallet::weight((1000, DispatchClass::Operational,))]
+		pub fn set_meetup_time_offset(
+			origin: OriginFor<T>,
+			meetup_time_offset: T::Moment,
+		) -> DispatchResultWithPostInfo {
+			<T as pallet::Config>::CeremonyMaster::ensure_origin(origin)?;
+			<MeetupTimeOffset<T>>::put(meetup_time_offset);
```
brenzi, line 421: here... only during REGISTERING

pifragile: 👍🏽

pifragile: done

brenzi (approved):

PR #166, opened 2022-03-07
[balances/#56] benchmark balances pallet

pifragile:
benchmarks can be tested without building the node by running

```cargo test --features runtime-benchmarks --manifest-path balances/Cargo.toml -- benchmarking```

from the `pallets` repo root.

### DIFF
--- a/balances/Cargo.toml
+++ b/balances/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["encointer.org <alain@encointer.org> and Parity Technologies <admin@p
 edition = "2021"
 
 [dependencies]
+approx = "0.5.0"
 codec = { package = "parity-scale-codec", version = "2.3.1", default-features = false, features = ["derive"] }
 log = { version = "0.4.14", default-features = false }
 scale-info = { version = "1.0", default-features = false }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: balances/Cargo.toml [hunk 1]
```diff
@@ -5,6 +5,7 @@ authors = ["encointer.org <alain@encointer.org> and Parity Technologies <admin@p
 edition = "2021"
 
 [dependencies]
+approx = "0.5.0"
 codec = { package = "parity-scale-codec", version = "2.3.1", default-features = false, features = ["derive"] }
```
clangenb, line 9: As `approx` is only used in the benchmarks, it should be an optional dependency and enable in with the `benchmark` feature.

balances/src/benchmarking.rs
```diff
@@ -0,0 +1,32 @@
...
+	balances::{BalanceType, Demurrage},
+	fixed::traits::LossyInto,
+};
+use frame_benchmarking::{account, benchmarks, impl_benchmark_test_suite};
+use frame_support::parameter_types;
+use frame_system::RawOrigin;
+
+parameter_types! {
+	pub const DefaultDemurrage: Demurrage = Demurrage::from_bits(0x0000000000000000000001E3F0A8A973_i128);
+}
```
clangenb, line 13: What do you need this for?

clangenb (commented):
Looks good in general, only minor comments.

balances/src/benchmarking.rs
pifragile: nothing😂

clangenb (approved):
Looks good to me

brenzi (commented):
you call the benchmark test `transfer` but you are actually benchmarking `issue`. shouldn't that be two distinct benchmarks?

clangenb:
No, issue is still in the setup phase, the benchmarked code is this line:

`_(RawOrigin::Signed(alice.clone()), bob_lookup, cid, BalanceType::from_num(10i32))`

the `_` is repblaced with the benchmark name, and will therefore execute the `transfer` dispatchable.

brenzi (approved):

PR #170, opened 2022-03-09
`ceremonies_getReputation` returns now `CeremonyIndex` too

clangenb:
The RPC did not return meaningful values before. We could not infer the ceremony index the reputations were valid.
 
Now the return value has been changed to: `Vec<(CeremonyIndex, CommunityReputation)>`, where `CommunityReputation = { community_identifier, reputation }`.


brenzi (approved):
LGTM

PR #180, opened 2022-03-15
[ceremonies/#56] benchmark ceremonies pallet

pifragile:


ceremonies/Cargo.toml
```diff
@@ -17,11 +17,13 @@ encointer-communities = { package = "pallet-encointer-communities", path = "../c
 encointer-scheduler = { package = "pallet-encointer-scheduler", path = "../scheduler", default-features = false }
 
 # substrate deps
+frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, optional = true, branch = "master" }
 frame-system = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 frame-support = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-std = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
+sp-core = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "master" }
```
clangenb, line 26: Default features must be false. Otherwise, no-std builds are broken. Also, this should probably be optional because it is only used in benchmarks?

ceremonies/src/benchmarking.rs
```diff
@@ -0,0 +1,386 @@
...
+		where
+		<T as frame_system::Config>::AccountId: From<sp_core::sr25519::Public>,
+		<T as Config>::Signature: From<sp_core::sr25519::Signature>,
+		<T as pallet_timestamp::Config>::Moment: From<u64>,
+		<T as frame_system::Config>::BlockNumber: From<u64>,
+		<T as frame_system::Config>::Event: From<pallet::Event<T>>
+	}
+
+	// register reputable
+	register_participant {
```
clangenb, line 214: Did you decide to cover multiple cases here because you don't know which one is the longest path?

ceremonies/src/benchmarking.rs
```diff
@@ -0,0 +1,386 @@
...
+		blocknr += 1.into();
+		run_to_block::<T>(blocknr);
+	}
+}
+
+pub fn set_timestamp<T: Config>(t: T::Moment) {
+	let _ = pallet_timestamp::Pallet::<T>::set(RawOrigin::None.into(), t);
+}
+
+fn get_proof_of_attendance<T: Config>(
```
clangenb, line 164: I know this is copied code, but the naming confuses me.

* `get_proof_of_attendance`: I'd expect to only get the proof, not actually inserting the fake reputation into the runtime
* `prove_attendance`: This one I'd rather expect to change state in the runtime instead of just returning the proof object.

Please change:
* rename: `get_proof_of_attendance` -> `fake_last_attendance_and_get_proof`
* rename: `prove_attendance`: `create_proof_of_attendance`

I happily accept other suggestions.

ceremonies/src/benchmarking.rs
clangenb: I am quite confident that all the paths are of similar complexity. We should only go for the worst-case scenario instead of introducing more complexity here.

ceremonies/src/benchmarking.rs
```diff
@@ -0,0 +1,386 @@
...
+			let p = users[i as usize].clone();
+			if i < 8 {
+			assert_ok!(Pallet::<T>::register_participant(RawOrigin::Signed(account_id::<T>(&(p.clone()))).into(), cid, Some(proofs[i as usize].clone())));
+			} else {
+				assert_ok!(Pallet::<T>::register_participant(RawOrigin::Signed(account_id::<T>(&(p.clone()))).into(), cid, None));
+			}
+		}
+
+		run_to_next_phase::<T>();
+		run_to_next_phase::<T>();
```
clangenb, line 335: This is very redundant with the setup code from the `attest_claims` benchmark. I guess you can easily extract some functions here.

clangenb (commented):
Looks in general quite good, I have some remarks though. :)

ceremonies/src/benchmarking.rs
```diff
@@ -0,0 +1,377 @@
...
+				Some(proofs[i as usize].clone())
+			));
+		} else {
+			assert_ok!(Pallet::<T>::register_participant(
+				RawOrigin::Signed(account_id::<T>(&(p.clone()))).into(),
+				cid,
+				None
+			));
+		}
+	}
```
clangenb, line 241: This can be a bit more rust ergonomic:

* & is a borrow; hence you don't need ownership here, so a clone does nothing. I think it is more ergonomic like this:

```
// iter() returns &Pair, into_iter() return Pair (owned value)
for (i, p) users.iter().enumerate() {
    let maybe_proof = None
    if i < num_reputables {
	 maybe_proof = Some(proofs[i].clone())
    }
        
    assert_ok!(Pallet::<T>::register_participant(
	              RawOrigin::Signed(account_id::<T>(p)).into(),
		      cid,
		      maybe_proof
	      ));
}
```

ceremonies/src/benchmarking.rs
```diff
@@ -0,0 +1,377 @@
...
+	let mut users: Vec<sr25519::Pair> = vec![];
+	let mut proofs: Vec<ProofOfAttendance<T::Signature, T::AccountId>> = vec![];
+
+	let num_users_total = num_newbies + num_reputables;
+	// create users and fake reputation
+	for i in 0..num_users_total {
+		let p = sr25519::Pair::from_entropy(&[i as u8; 32], None).0;
+		users.push(p.clone());
+		if i < num_reputables {
+			proofs.push(fake_last_attendance_and_get_proof::<T>(&p.clone(), cid));
```
clangenb, line 221: Just `&p` is enough

ceremonies/src/benchmarking.rs
```diff
@@ -0,0 +1,377 @@
...
+	let community_meta: CommunityMetadataType = CommunityMetadataType {
+		name: "Default".into(),
+		symbol: "DEF".into(),
+		..Default::default()
+	};
+	encointer_communities::Pallet::<T>::new_community(
+		RawOrigin::Root.into(),
+		location,
+		bs.clone(),
+		community_meta.clone(),
```
clangenb, line 35: I think creating the metadata is unnecessary, you should just be able to create it in place. (You would not need to clone anyhow, as you don't use the metadata afterwards;

```suggestion
		CommunityMetadata::Default(),
```

clangenb (commented):
Much better now, but I still have some comments. :)

ceremonies/src/benchmarking.rs
pifragile: sexy :)

clangenb (approved):
Great, looks good to me now.

brenzi (approved):

brenzi:
@pifragile sorry, merged a hotfix and it seems your tests are not working after merge. didn't look into it though

pifragile:
> @pifragile sorry, merged a hotfix and it seems your tests are not working after merge. didn't look into it though

@brenzi ok, will check

brenzi:
there is a new trivial extrinsic that may need to be included in benchmarks

pifragile:
> there is a new trivial extrinsic that may need to be included in benchmarks

pifragile:
> there is a new trivial extrinsic that may need to be included in benchmarks

@brenzi ok fixed failing tests and added a benchmark for `purge_community_ceremony`

PR #181, opened 2022-03-15
Signed meetup offset

### LINKED ISSUE #149: implement signed meetup time offset to tune or vary the exact daytime relative to high sun

pifragile:
* Closes #149

### DIFF
--- a/ceremonies/assignment/src/lib.rs
+++ b/ceremonies/assignment/src/lib.rs
@@ -4,8 +4,8 @@
 
 use crate::math::{checked_ceil_division, checked_modulo, find_prime_below, mod_inv};
 use encointer_primitives::{
-	ceremonies::{AssignmentParams, MeetupIndexType, ParticipantIndexType},
-	communities::{Degree, Location, LossyFrom},
+	ceremonies::{AssignmentParams, MeetupIndexType, MeetupTimeOffsetType, ParticipantIndexType},
+	communities::{Location, LossyFrom},
 	RandomNumberGenerator,
 };
 use sp_runtime::{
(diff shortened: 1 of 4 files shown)


clangenb (approved):
Looks good to me.

clangenb (approved):

ceremonies/src/lib.rs
```diff
@@ -415,12 +415,18 @@ pub mod pallet {
...
-			meetup_time_offset: T::Moment,
+			meetup_time_offset: MeetupTimeOffsetType,
 		) -> DispatchResultWithPostInfo {
 			<T as pallet::Config>::CeremonyMaster::ensure_origin(origin)?;
 			if <encointer_scheduler::Pallet<T>>::current_phase() != CeremonyPhaseType::REGISTERING {
 				return Err(<Error<T>>::WrongPhaseForChangingMeetupTimeOffset.into())
 			}
+
+			// Meetup time offset needs to be in [-8h, 8h]
+			if meetup_time_offset.abs() > 8 * 3600 * 1000 {
```
brenzi, line 426: never hardcode arbitrary numbers! please make this a constant (doesn't need to be a configurable runtime constant, just a `const`)

ceremonies/assignment/src/lib.rs
```diff
@@ -173,19 +173,21 @@ pub fn meetup_time<Moment: Copy + AtLeast32Bit>(
...
+	let lon: i64 = (i64::lossy_from(location.lon.round_to_zero()) - 180i64).abs();
 
-	if location.lon < Degree::from_num(0) {
-		attesting_start + one_day / Moment::from(2u32) + lon_time_abs + offset
-	} else {
-		attesting_start + one_day / Moment::from(2u32) - lon_time_abs + offset
-	}
+	let lon_time = lon * per_degree;
+	let attesting_start_u64: u64 = attesting_start.saturated_into();
+	let meetup_time = attesting_start_u64 as i64 + lon_time + offset as i64;
```
brenzi, line 189: so much nicer, thx!

brenzi (changes requested):
very nice. just one tiny thing commented

ceremonies/src/lib.rs
brenzi: well, actually....forget it. in this case we should prevent that someone thinks this could be used to shift by more than 12h, which would break

brenzi (approved):
changed my mind about hardcoding in this case

ceremonies/assignment/src/lib.rs
pifragile: ✌🏽☺️

PR #184, opened 2022-03-16
implement reputation caching for rpc

### LINKED ISSUE #118: optimize performance of reputation rpc getter
Currently we have a very expensive implementation (#113) for getting all reputation history for one AccountId. 

Elsewhere, we use offchain indexing to allow more performant cached queries:

https://github.com/encointer/pallets/blob/master/communities/rpc/src/lib.rs
cache update triggered by:
https://github.com/encointer/pallets/blob/e3c4d7fc9a2ca66f8589842a9ef432bf4129ef7c/communities/src/lib.rs#L135

For this we should invert the mapping of the storage in the cache: from 
https://github.com/encointer/pallets/blob/e3c4d7fc9a2ca66f8589842a9ef432bf4129ef7c/ceremonies/src/lib.rs#L101
(which is correct for the runtime pallet, because we need purge by cindex to be cheap)
to:
```
AccountId-> Vec<(Community, Reputation)>
```
because the rpc will query by account, not by cindex or community.

The cache should always return a vec of length (ReputationLifetime) with tuples of Community and Reputation. The tuple makes sense if that accountid has participated in more than one community in the recent ceremonies (and we're sure by design that one accountid can't participate in two communities for the same ceremony....however, technically, we should handle that collision somehow) 

Good thing about caching: the cache only gets dirty once per ceremony, so the refresh may be costly if we gain significant query performance benefits

pifragile:
* Closes #118


ceremonies/rpc/src/lib.rs
```diff
@@ -38,26 +45,90 @@ where
...
+	pub fn refresh_reputation_cache(
+		&self,
+		account: AccountId,
+		at: Option<<Block as BlockT>::Hash>,
+	) {
+		let api = self.client.runtime_api();
+		let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
+		let reputations =
+			api.get_reputations(&at, &account).map_err(runtime_error_into_rpc_err).unwrap();
+		let cache_key = &(STORAGE_REPUTATION_KEY, account.as_slice()).encode()[..];
```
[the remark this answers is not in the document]
clangenb: Actually, the cleanest approach would be to define a function in the primitives that derives the cache key because it is used multiple times, and it is very easy to introduce discrepancies.

```
pub fn reputation_cache_key<Account: Encode>(account: Account) -> Vec<u8> {
      (STORAGE_REPUTATION_KEY, account).encode()
}
```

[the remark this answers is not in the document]
clangenb: Please to the same for the cache dirty bit.

ceremonies/rpc/src/lib.rs
```diff
@@ -66,13 +137,27 @@ where
...
 
-		let api = self.client.runtime_api();
-		let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
-		return Ok(api.get_reputations(&at, &account).map_err(runtime_error_into_rpc_err)?)
+		if !self.offchain_indexing {
+			return Err(offchain_indexing_disabled_error("ceremonies_getReputations"))
+		}
+
+		if self.cache_dirty(
+			[REPUTATION_CACHE_DIRTY_KEY, account.encode().as_slice()].concat().as_slice(),
```
clangenb, line 145: Why do you derive the reputation cache key, and the dirty bit cache key in two different ways?

Anyhow, please use a function as I said above.

ceremonies/rpc/Cargo.toml
```diff
@@ -21,3 +21,6 @@ sc-rpc-api = { version = "0.10.0-dev", git = "https://github.com/paritytech/subs
 sp-api = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "master" }
 sp-blockchain = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "master" }
 sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "master" }
+
+[dev-dependencies]
+sp-core = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "master" }
```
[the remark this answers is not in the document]
pifragile: oh it was for the tests that i removed again

ceremonies/rpc/src/lib.rs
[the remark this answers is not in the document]
pifragile: when you explain it like this, it sounds so easy :) but it somehow isn't😅

ceremonies/rpc/src/lib.rs
```diff
@@ -66,13 +134,25 @@ where
...
-		return Ok(api.get_reputations(&at, &account).map_err(runtime_error_into_rpc_err)?)
+		if !self.offchain_indexing {
+			return Err(offchain_indexing_disabled_error("ceremonies_getReputations"))
+		}
+
+		if self.cache_dirty(&reputation_cache_dirty_key(account.encode())) {
+			self.refresh_reputation_cache(account.clone(), at);
+		}
+
+		let cache_key = &reputation_cache_key(account.encode());
```
clangenb, line 145: No need to use `.encode()` here, we don't need to double encode, just pass the account as is. Please remove that everywhere.

ceremonies/rpc/src/lib.rs
```diff
@@ -38,26 +45,87 @@ where
 	) -> Result<Vec<(CeremonyIndexType, CommunityReputation)>>;
 }
 
-pub struct Ceremonies<Client, Block, AccountId> {
+pub struct Ceremonies<Client, Block, AccountId, S> {
 	client: Arc<Client>,
 	_marker: std::marker::PhantomData<(Block, AccountId)>,
```
clangenb, line 50: Minor detail (I know this has been like this before). The `_marker` is usually the last field because it is never used.

clangenb (commented):
Some minor details, but then we are there.

ceremonies/rpc/src/lib.rs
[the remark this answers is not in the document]
clangenb: haha, it took me a while to get there :)

ceremonies/rpc/src/lib.rs
clangenb: I could imagine that you can even pass a `&AccountId` to the `reputation_cache_key` function, might be worth a try.

ceremonies/rpc/src/lib.rs
pifragile: i dont understand, please elaborate

ceremonies/rpc/src/lib.rs
pifragile: yep works

ceremonies/Cargo.toml
```diff
@@ -22,8 +22,7 @@ frame-support = { version = "4.0.0-dev", git = "https://github.com/paritytech/su
 pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-std = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
-
-# benchmarking
+sp-io = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, optional = true, branch = "master" }
```
clangenb, line 26: Minor thing, this is not an optional dependency for benchmarking, please put in the above section, in the suitable alphabetical entry.

Also, please add `sp-io/std` in the std feature section below; otherwise this may have unforeseen effects.

clangenb (commented):
Tiny remark, looks good otherwise 👍

ceremonies/Cargo.toml
pifragile: what do you mean by `the above section`?

ceremonies/Cargo.toml
```diff
@@ -22,8 +22,7 @@ frame-support = { version = "4.0.0-dev", git = "https://github.com/paritytech/su
 pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-std = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
```
clangenb, line 24: Above section == here.

Also, please re-introduce the deleted `# benchmarking` on line 26.

clangenb (approved):
LGTM

PR #187, opened 2022-03-19
Fixes for node-integration

clangenb:
This predominantly contains changes in the ceremonies-benchmark, which had multiple issues:

* `no_std` incompatibility due to crypto usage. This was circumvented by not using the `sp_core::crypto::sr25519` module. Instead, we created our own crypto application keys with `sp-application-crypto`. This approach is taken from [frame-benchmarking](https://github.com/paritytech/substrate/blob/master/frame/benchmarking/src/baseline.rs).
* the node disliked that we played with block numbers, timestamps and `on_finalize` hooks. Updating the phase was simply replaced by a `scheduler::next_phase` call.
* fixed some unrealistic trait bounds and cleaned up the benchmarking in general.

General Changes:
 * Fix some no_std stuff.

Tested:
* benchmarks run for all pallets

### DIFF
--- a/bazaar/src/benchmarking.rs
+++ b/bazaar/src/benchmarking.rs
@@ -4,13 +4,14 @@ use encointer_primitives::communities::{
 };
 use frame_benchmarking::{account, benchmarks, impl_benchmark_test_suite};
 use frame_system::RawOrigin;
+use sp_std::borrow::ToOwned;
 
-fn test_url() -> String {
-	"https://test.com".to_string()
+fn test_url() -> PalletString {
+	"https://test.com".to_owned().into()
 }
 
-fn example_url() -> String {
-	"https://example.com".to_string()
+fn example_url() -> PalletString {
+	"https://example.com".to_owned().into()
 }
 
 fn create_community<T: Config>() -> CommunityIdentifier {
--- a/primitives/src/ceremonies.rs
+++ b/primitives/src/ceremonies.rs
@@ -37,9 +37,6 @@ pub type MeetupTimeOffsetType = i32;
 #[cfg(not(feature = "std"))]
 use sp_std::vec::Vec;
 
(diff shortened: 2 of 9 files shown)


bazaar/src/benchmarking.rs
```diff
@@ -4,13 +4,14 @@ use encointer_primitives::communities::{
 };
 use frame_benchmarking::{account, benchmarks, impl_benchmark_test_suite};
 use frame_system::RawOrigin;
+use sp_std::borrow::ToOwned;
 
-fn test_url() -> String {
-	"https://test.com".to_string()
+fn test_url() -> PalletString {
+	"https://test.com".to_owned().into()
```
clangenb, line 10: String is not-available in the runtime.

primitives/src/ceremonies.rs
```diff
@@ -161,10 +158,10 @@ impl<Signature, AccountId: Clone + Encode, Moment: Encode + Copy>
 			.encode()
 	}
 
-	#[cfg(feature = "std")]
+	#[cfg(any(feature = "std", feature = "full_crypto"))]
 	pub fn sign<P>(self, pair: &P) -> Self
```
clangenb, line 162: Not needed now actually, but might be useful, if we ever migrate to the enclave ;)

test-utils/src/lib.rs
```diff
@@ -122,7 +122,7 @@ macro_rules! impl_frame_system {
 
 pub type Moment = u64;
 parameter_types! {
-	pub const MinimumPeriod: Moment = 1;
+	pub const MinimumPeriod: Moment = BLOCKTIME / 2;
```
clangenb, line 125: Changed this while trying another approach, but I thought why not use realistic numbers.

brenzi (approved):

PR #190, opened 2022-03-30
Fees in community currency

pifragile:
Fot the node integration see https://github.com/encointer/encointer-node/pull/194


balances/src/lib.rs
```diff
@@ -46,6 +58,7 @@ pub mod pallet {
 	use super::*;
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
+	use sp_runtime::AccountId32;
```
clangenb, line 61: We should not use the explicit `AccountId` in the pallet. Always refer to `<<T as frame_system>::Config>::AccountId`

balances/src/impl_fungibles.rs
```diff
@@ -0,0 +1,180 @@
...
+		asset: <Pallet<T> as fungibles::Inspect<T::AccountId>>::AssetId,
+		fungible_balance: <Pallet<T> as fungibles::Inspect<T::AccountId>>::Balance,
+	) -> BalanceType {
+		let decimals = Self::decimals(&asset);
+		let mut result: BalanceType = BalanceType::from_num(0);
+
+		result = result +
+			BalanceType::from_num(
+				((fungible_balance << 64) >> 64) as f64 / (10i128.pow(decimals as u32) as f64),
+			);
```
clangenb, line 59: `((fungible_balance << 64) >> 64)` I guess this is to truncate, we might find some higher level function for this. :)

communities/src/lib.rs
```diff
@@ -696,6 +701,83 @@ impl<T: Config> Pallet<T> {
 	}
 }
 
+pub type OnChargeTransactionOf<T> = <T as pallet_transaction_payment::Config>::OnChargeTransaction;
```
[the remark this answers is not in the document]
clangenb: Ok, I see because of the dependency stuff, hence I propose to define this thing in a separate crate.

balances/src/impl_fungibles.rs
```diff
@@ -0,0 +1,180 @@
...
+				BalanceType::from_num(
+					2i128.pow(64) as f64 / 10i128.pow(decimals as u32) as f64,
+				);
+		result
+	}
+}
+
+impl<T: Config> fungibles::Inspect<T::AccountId> for Pallet<T> {
+	type AssetId = CommunityIdentifier;
+	type Balance = u128;
```
clangenb, line 72: I think this is a fundamental misunderstanding. I think what you did works, but is much too complicated. We should use `BalanceType = EncointerBalanceType` here, so we can spare all those conversions.

balances-tx-payment-extension/src/balance_conversion.rs
```diff
@@ -0,0 +1,88 @@
...
+	let conversion_factor = ((0.01f64 / 5.233f64) *
+		10i128.pow((asset_balance_decimals - 10) as u32) as f64 *
+		fee_conversion_factor as f64) as u128;
+
+	// assuming a nominal income of 20
+	return (balance * conversion_factor * reward) /
+		BalancesPallet::<T>::balance_type_to_fungible_balance(
+			cid.into(),
+			BalanceType::from_num(20i32),
+		)
```
clangenb, line 46: I think this calculation is wrong, and should not incorporate any hardcoded numbers. Just do it as @brenzi said it should be:

```
CCfee = KSMfee * factor * reward(cid)
```

communities/Cargo.toml
```diff
@@ -21,6 +21,8 @@ frame-system = { version = "4.0.0-dev", git = "https://github.com/paritytech/sub
 sp-io = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-std = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
 sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "master" }
+pallet-asset-tx-payment  = { version = "4.0.0-dev" , default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+pallet-transaction-payment = { version = "4.0.0-dev" , default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
```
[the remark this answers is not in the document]
clangenb: resolved as it is no longer in this crate anyhow.

balances/src/impl_fungibles.rs
clangenb: Ok, I see. This is not possible, it needs to be a scalar type implementing `AtLeast32Bit`.

primitives/src/balances.rs
```diff
@@ -48,3 +51,46 @@ pub struct BalanceEntry<BlockNumber> {
...
+}
+
+impl Convert<u128, BalanceType> for EncointerBalanceConverter {
+	fn convert(fungible_balance: u128) -> BalanceType {
+		let decimals = ENCOINTER_BALANCE_DECIMALS;
+
+		let mut result: BalanceType = BalanceType::from_num(0);
+
+		result += BalanceType::from_num(
+			((fungible_balance << 64) >> 64) as f64 / (10i128.pow(decimals as u32) as f64),
```
clangenb, line 88: These conversions might happen in the runtime. You must never use indeterministic data types there.

balances/src/impl_fungibles.rs
clangenb: This is actually the same as `fungible_balance as u64`

balances-tx-payment-extension/src/balance_conversion.rs
clangenb: I updated the meaning of the `FeeConversionFactor`

primitives/src/balances.rs
clangenb: Changed to using fixed piont types.

clangenb (approved):
Although I had quite a few comments, it looked good in general. I just resolved all the remarks myself due to time-constraints. Don't take it personally. :)

Just a few notes:
* never use floating-point types by stuff which is executed in the runtime. This is not deterministic and will lead to consensus errors.
* you sometimes use quite crazy math ops, although they might be very elegant, they hard to follow. Some more doc there would be nice.

Ready to merge from my point of view. It compiles with the node.

balances-tx-payment-extension/src/balance_conversion.rs
brenzi: One consideration here: whatever the formula, we should ensure that we have enough slack in both directions to adjust fees. I.e. if we define the factor as an u64 and initialize it to a value of 5 that would be bad because we can only lower fees by a factor of 5. Rather add a few zeros in that case

balances-tx-payment/src/balance_conversion.rs
```diff
@@ -0,0 +1,65 @@
...
+/// Transforms the native token to the community currency
+///
+/// Assumptions:
+/// * Native token has 12 decimals
+/// * fee_conversion_factor is in Units 1 / [KSM]
+pub fn balance_to_community_balance(
+	balance: u128, // Unit = [pKSM]
+	reward: u128,
+	fee_conversion_factor: u32,
+) -> u128 {
```
brenzi, line 34: Please comment how a u128 will be interpreted as a CC balance U64F64
Like this it is difficultbto interpret the formula

balances/src/impl_fungibles.rs
```diff
@@ -0,0 +1,180 @@
+use super::*;
```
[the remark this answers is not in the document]
pifragile: i just tried to implement the smallest subset necessary for our usecase :)

balances/src/impl_fungibles.rs
pifragile: the problem was that it needed to be unsigned

balances-tx-payment-extension/src/balance_conversion.rs
pifragile: i set the factor to 10000 in the genesis config and calculated the formula based on this factor and the constraints discussed with @brenzi . i think 10000 allows for quite some slack

balances-tx-payment/src/tests.rs
```diff
@@ -1,21 +1,20 @@
-use crate::balance_to_community_balance;
+use crate::{balance_to_community_balance, ONE_MICRO_KSM};
 use rstest::*;
 
-/// 1 micro KSM with 12 decimals
-const ONE_MICRO_KSM: u128 = 1_000_000;
-
 /// 1 community currency token with 18 decimals
 const ONE_CC: u128 = 1_000_000_000_000_000_000;
```
brenzi, line 5: I don't get this. If we have BalanceType U64F64 this means one cc is `1 << 64` not `1_000_000_000_000_000_000`
This needs more explaining comments, even if the new constants already improve readability

brenzi: @pifragile Why did you arbitrarily chose to set ONE_CC=1_000_000_000_000_000_000? I get that you tried to pick a number that avoids overflow, but your choice has nothing to do with the real interpretation of "one CC". We don't need a power of 10, we need the "real" ONE_CC. 

Or: why didn't you just set `ONE_CC: u128 = 1 << 64`? Then we could typecast this value to U64F64 and would really get 1 CC

pifragile: @brenzi the rationale behind this is that i assumed that fungibles uses a system similar to Ethereum where numbers are represented as a large integer (like 256 or 128 bit) and a given amount of decimals.
See here: https://github.com/paritytech/substrate/blob/ded44948e2d5a398abcb4e342b0513cb690961bb/frame/support/src/traits/tokens/fungibles/metadata.rs#L29

So for example with 5 decimals, the number `x` will be represented as `x * 10^5`. I chose 18 decimals, hence this number.

brenzi: Ok, but that confirms my concern. We use fixpoint with a base of 2, not base of 10. Unless you can convince me otherwise, I'd like this changed to 
`ONE_CC: u128 = 1 << 64` throughout

brenzi (changes requested):
as requested in comments

balances-tx-payment/src/tests.rs
pifragile: this only works if no client of the balances pallet ever uses the `decimals` field, because we will obviously not be able to put a value in that field if we use your apporach. i will do some research. what is the reason you would like to change this? to avoid the conversions between base 2 / base 64 numbers?

balances-tx-payment/src/balance_conversion.rs
pifragile: in this step there is actually no U64F64 involved. we convert between a `u128` KSM value and a `u128` fungible balance

brenzi (approved):

PR #193, opened 2022-04-02
Add missing storage setters

### LINKED ISSUE #191: setters for time_tolerance and location_tolerance
https://github.com/encointer/pallets/blob/74a57e098dd7e9a9f6f54147a90158ab216d8f60/ceremonies/src/lib.rs#L820-L821

we forgot these. currently, we need sudo to set these with set_storage. But CeremonyMaster should be allowed to set these

brenzi:
closes #191 
* fixes hardcoded dummy weight for set_meetup_offset

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -448,7 +448,7 @@ pub mod pallet {
 			Ok(().into())
 		}
 
-		#[pallet::weight((1000, DispatchClass::Operational,))]
+		#[pallet::weight((<T as Config>::WeightInfo::set_time_tolerance(), DispatchClass::Normal, Pays::Yes))]
 		pub fn set_time_tolerance(
 			origin: OriginFor<T>,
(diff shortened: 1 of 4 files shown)


ceremonies/src/lib.rs
```diff
@@ -427,7 +427,7 @@ pub mod pallet {
 			Ok(().into())
 		}
 
-		#[pallet::weight((1000, DispatchClass::Operational,))]
+		#[pallet::weight((<T as Config>::WeightInfo::set_meetup_time_offset(), DispatchClass::Normal, Pays::Yes))]
```
brenzi, line 430: don't know how criticsal this is, but such errors could be detected by some custom check script we run in CI to ensure no hardcoded weights are present

clangenb: Yeah makes sense. `1000` is very little and would make an attack quite cheap.

clangenb (approved):
LGTM

PR #196, opened 2022-04-05
Aggregated account data

### LINKED ISSUE #157: Custom RPC: Aggregated Account Data
Let client get all encointer-relevant state data for one account and community 
for a mobile app, along the notes in encointer/encointer-wallet-flutter#468.

The format is taken from the above issue:

```
{
  "global": {
    "ceremony_phase": "REGISTERING",
    "ceremony_index": 42
  },
  "personal": {
    "participant_type": "NEWBIE",
    "participant_index": 32,
    "meetup_index": 2,
    "meetup_location_index": 33,
    "meetup_time": "2022-03-15 12:50",
    "meetup_registry": [
      "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
      "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
      "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
      "5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy",
      "5HGjWAeFDfFCWPsjFQdVV2Msvz2XtMktvgocEZcCj68kUMaw",
      "5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL"
    ]
  }
}
```

With the sole difference that I removed the part about the balances. I think this one should be separate with the reasoning:

* All fields are no very stationary. They are only updated upon phase changes, or when a participant registers herself. (Might change if endorsements are going to be allowed after registration.) So no need to subscribe, we just fetch all de data occasionally.
* Now this only needs access to the ceremonies-pallet, and should be fairly easy to implement.
* It still solved both core issues we try to solve: complex dependencies, and the introduced latency thereof.

The only thing we need to check:
* Can we access the runtime's storage value directly from the rpc (the host)? (I think so). If not, we'd need to introduce a runtime-api call. Similar to how we get the reputations for an account.

pifragile:
* Closes #157


ceremonies/src/lib.rs
```diff
@@ -899,6 +899,63 @@ impl<T: Config> Pallet<T> {
...
+				let locations = <encointer_communities::Pallet<T>>::get_locations(&cid);
+				let location_assignment_params = Self::assignments((cid, cindex)).locations;
+
+				meetup_location_index = get_meetup_location_index(
+					participant_meetup_index,
+					&locations,
+					location_assignment_params,
+				);
+				let location =
+					Self::get_meetup_location((cid, cindex), participant_meetup_index).unwrap();
```
clangenb, line 935: No unwraps in runtime code. If there is a formal proof that this does not panic, use `expect("proof; qed");`

ceremonies/src/lib.rs
```diff
@@ -899,6 +899,63 @@ impl<T: Config> Pallet<T> {
...
+			AggregatedAccountDataPersonal<T::AccountId, T::Moment>,
+		>;
+
+		// check if the participant is registered. if not the entire personal field will be None.
+		if let Some(participant_type) = Self::get_participant_type((cid, cindex), account) {
+			let mut meetup_location_index: Option<MeetupIndexType> = None;
+			let mut meetup_time: Option<T::Moment> = None;
+			let mut meetup_registry: Option<Vec<T::AccountId>> = None;
+			let mut meetup_index: Option<MeetupIndexType> = None;
+
```
clangenb, line 922: Why do you predefine those fields, we same some lines of code if this is defined in place.

ceremonies/src/tests.rs
```diff
@@ -1804,3 +1807,134 @@ fn set_location_tolerance_works() {
...
+			EncointerCeremonies::get_aggregated_account_data(cid, &bootstrapper);
+
+		assert_eq!(aggregated_account_data.global.ceremony_phase, CeremonyPhaseType::REGISTERING);
+		assert_eq!(aggregated_account_data.global.ceremony_index, 1);
+		let mut personal = aggregated_account_data.personal.unwrap();
+		assert_eq!(personal.participant_type, ParticipantType::Bootstrapper);
+		assert_eq!(personal.meetup_index, None);
+		assert_eq!(personal.meetup_location_index, None);
+		assert_eq!(personal.meetup_time, None);
+		assert_eq!(personal.meetup_registry, None);
```
clangenb, line 1902: I don't really understand this case, how can this happen the way I read the code this should fail. Can you help me here? :)

clangenb (commented):
Looks good, I also like the refactoring you started, (which created space for even more refactoring, which I'd like to see :))

clangenb:
@brenzi I only have code-quality remarks. I think from the functionality it looks fine, so from my point of view we can merge this to get it into the parachain.

brenzi:
well, an unwarp() in the runtime is not only a code quality problem. It is serious. but it's not part of an extrinsic execution, only rpc, so not strictly critical

ceremonies/src/tests.rs
pifragile: it is the case when the participant is registered, but the assignment has not yet been made

ceremonies/src/lib.rs
```diff
@@ -1189,53 +1246,87 @@ impl<T: Config> Pallet<T> {
...
+						Self::bootstrapper_index(community_ceremony, &participant) - 1;
+					if participant_index < assignment_count.bootstrappers {
+						return meetup_index(
+							participant_index,
+							assignment.bootstrappers_reputables,
+							meetup_count,
+						)
+					}
+					None
+				},
```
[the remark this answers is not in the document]
pifragile: done. moving to a function would not make sense because you would also have to pass `assignment` and `assignment_count` to the function which would literally move all the functionality to the other function which seems pointless to me.

ceremonies/src/lib.rs
pifragile: i do it because i use a bunch of if statements to assign those variables. if i dont predefine them here, they woule be undefined in some cases. if there is a more elegant solution to this, please give me a hint :)

brenzi:
Thanks for your very modular fixes. That makes it a bliss to review!

ceremonies/src/tests.rs
clangenb: Ah, ok that makes sense!

PR #197, opened 2022-04-05
fix inverse assignment bug

### LINKED ISSUE #194: Fix unexpected behaviour if only one participant of a category is registered
There seems to be a bug in:
https://github.com/encointer/pallets/blob/80442d3f3f89cb0b6661cebb4f8c5706f2f1556b/ceremonies/assignment/src/lib.rs#L124

The subtraction overflows in the following configuration: 2 bootstrappers, 2 reputables, 1 newbie.

I assume that the problem is that fir the newbies, `participant_count` is 1 and `assignment.params.m` is 2, because 2 is the minimum possible value.

One should introduce a reasonable check for this case

pifragile:
* Closes #194

### DIFF
--- a/ceremonies/assignment/src/lib.rs
+++ b/ceremonies/assignment/src/lib.rs
@@ -121,7 +121,8 @@ pub fn assignment_fn_inverse(
 		}
 
 		result.push(t3);
-		if t3 < participant_count - assignment_params.m {
+
+		if t3 + assignment_params.m < participant_count {
 			result.push(t3 + assignment_params.m)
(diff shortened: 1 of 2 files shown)


clangenb (commented):
Looks good, nice trice trick to solve the subtraction underflow.

If you feel like it, I'd love to see this test being parametrized with `rstests`. :)

clangenb (approved):

clangenb:
Will merge, there are more important things to do currently.

PR #202, opened 2022-04-06
Endorsement after registration

### LINKED ISSUE #117: Allow endorsement after participant registered
Currently, the endorsement needs to be registered by the bootstrapper before the endorsee registers. while this is the expected sequence as newbies habve zero balance to pay fees without an invite or endorsement, it could still make sense to allow an a-posteriori upgrade.

What should be done: Allow this flow:
1. Participant A registers as newbie
2. Some bootstrapper endorses A
3. participant A registers again and the registration for A gets upgraded from newbie to endorsee

Not sure if this causes trouble because participant indexes may change when removing a newbie from the registry.....

pifragile:
* Closes #117

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -572,6 +572,8 @@ pub mod pallet {
 		InvalidMeetupTimeOffset,
 		/// the history for given ceremony index and community has been purged
 		CommunityCeremonyHistoryPurged,
+		/// Unregistering can only be performed during the registering phase
+		WrongPhaseForUnregistering,
 	}
 
 	#[pallet::storage]
@@ -942,6 +944,45 @@ impl<T: Config> Pallet<T> {
 				ParticipantType::Newbie
 			};
 		Ok(participant_type)
+	}
+
+	/// removes a participant from the registry maintaining continuous indices
+	fn unregister_newbie(
+		cid: CommunityIdentifier,
+		cindex: CeremonyIndexType,
+		participant: &T::AccountId,
+	) -> Result<(), Error<T>> {
+		if <encointer_scheduler::Pallet<T>>::current_phase() != CeremonyPhaseType::REGISTERING {
+			return Err(<Error<T>>::WrongPhaseForUnregistering.into())
+		}
+
+		if !<NewbieIndex<T>>::contains_key((cid, cindex), &participant) {
+			return Ok(())
+		}
+
+		let participant_index = <NewbieIndex<T>>::get((cid, cindex), &participant);
+		let participant_count = <NewbieCount<T>>::get((cid, cindex));
+
+		<NewbieRegistry<T>>::remove((cid, cindex), &participant_index);
+		<NewbieIndex<T>>::remove((cid, cindex), &participant);
+
+		for idx in (participant_index + 1)..=participant_count {
+			let p = <NewbieRegistry<T>>::get((cid, cindex), &idx)
+				.expect("idx <= participant_count, so p is registered; qed");
+
+			let new_idx = idx - 1;
+			<NewbieIndex<T>>::insert((cid, cindex), &p, new_idx);
+			<NewbieRegistry<T>>::remove((cid, cindex), &idx);
+			<NewbieRegistry<T>>::insert((cid, cindex), &new_idx, &p);
+		}
+		<NewbieCount<T>>::insert(
+			(cid, cindex),
+			participant_count
+				.checked_sub(1)
+				.expect("participant is registered, thus participant_count > 0; qed"),
+		);
+
+		Ok(())
 	}
 
 	fn is_registered(
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -366,6 +366,11 @@ pub mod pallet {
 			<Endorsees<T>>::insert((cid, cindex), newbie.clone(), ());
 			<EndorseesCount<T>>::mutate((cid, cindex), |c| *c += 1); // safe; limited by AMOUNT_NEWBIE_TICKETS
 
+			if <NewbieIndex<T>>::contains_key((cid, cindex), &newbie) {
+				Self::unregister_newbie(cid, cindex, &newbie)?;
+				Self::register(cid, cindex, &newbie, false)?;
+			}
+
 			debug!(target: LOG, "bootstrapper {:?} endorsed newbie: {:?}", sender, newbie);
 			Self::deposit_event(Event::EndorsedParticipant(cid, sender, newbie));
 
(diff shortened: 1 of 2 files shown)


ceremonies/src/lib.rs
```diff
@@ -944,6 +951,45 @@ impl<T: Config> Pallet<T> {
...
+			<NewbieIndex<T>>::insert((cid, cindex), &p, new_idx);
+			<NewbieRegistry<T>>::remove((cid, cindex), &idx);
+			<NewbieRegistry<T>>::insert((cid, cindex), &new_idx, &p);
+		}
+		<NewbieCount<T>>::insert(
+			(cid, cindex),
+			participant_count
+				.checked_sub(1)
+				.expect("participant is registered, thus participant_count > 0; qed"),
+		);
```
clangenb, line 988: Why this for-loop? Is there a reason preventing us replacing the missing index with the participant with the highest index? This would be way more efficient.

clangenb: You could use the built-in storage-map method [swap](https://github.com/paritytech/substrate/blob/26eff96fed52fb2001f37f17e743939b0e2384bb/frame/support/src/storage/generator/double_map.rs#L165) to first swap the index with the last entry and then remove the last entry.

clangenb (changes requested):
I like the idea of directly mutating the participant to an `endorsee` upon endorsement.

But I believe you can do the process much more efficient.

ceremonies/src/lib.rs
pifragile: this would indeed be a lot more efficient. but it would not keep the order of registration which would be unfair in the case where not all newbies get assigned to a meetup.

pifragile: but we could argue that nobody gets a disadvantage, just one participant gets an advantage and this could be acceptable for the big performance gain...

brenzi: a for loop is inacceptable anyway as it is O(n) where n is the number of newbies which is unbounded

brenzi: can the system be played? is it predictable who gets the advantage from endorsing a newbie a posteriori?

pifragile: yes if there are `n` newbies with indices `1..=n`, and newbie with index `i` gets endorsed, then newbie with index `n` will receive index `i`

pifragile: but as a regular user in order to play the system, you would have to know exactly when the endorsement will happen

brenzi: I think it's acceptable. not nice, but much better than O(n)

ceremonies/src/lib.rs
```diff
@@ -944,6 +951,42 @@ impl<T: Config> Pallet<T> {
...
+		let participant_count = <NewbieCount<T>>::get((cid, cindex));
+
+		let last_participant = <NewbieRegistry<T>>::get((cid, cindex), &participant_count)
+			.expect("Indices are continuous, thus index participant_count is Some; qed");
+
+		<NewbieRegistry<T>>::insert((cid, cindex), &participant_index, &last_participant);
+		<NewbieIndex<T>>::insert((cid, cindex), &last_participant, &participant_index);
+
+		<NewbieRegistry<T>>::remove((cid, cindex), &participant_count);
+		<NewbieIndex<T>>::remove((cid, cindex), &participant);
```
clangenb, line 978: I think you are manually implementing the swap method I recommended in the comments before. Is there a reason for that?

clangenb: Ah yes, you are right, you actually make it better than what I proposed XD

clangenb (approved):
Looks good to me!

brenzi:
Please resolve merge conflicts, then we can merge

ceremonies/src/lib.rs
```diff
@@ -1007,6 +1014,42 @@ impl<T: Config> Pallet<T> {
...
+
+		if !<NewbieIndex<T>>::contains_key((cid, cindex), &participant) {
+			return Ok(())
+		}
+
+		let participant_index = <NewbieIndex<T>>::get((cid, cindex), &participant);
+		let participant_count = <NewbieCount<T>>::get((cid, cindex));
+
+		let last_participant = <NewbieRegistry<T>>::get((cid, cindex), &participant_count)
+			.expect("Indices are continuous, thus index participant_count is Some; qed");
```
brenzi, line 1035: hmmm, what if participant_count is zero?

brenzi: or: what if the newbie is the only entry? and participant_count is therefore 1?

brenzi: hmmm, gues then it would return above already....

brenzi (changes requested):

ceremonies/src/tests.rs
```diff
@@ -1071,6 +1071,32 @@ fn endorsing_two_newbies_works() {
...
+		assert!(NewbieIndex::<TestRuntime>::contains_key((cid, cindex), &yran));
+
+		assert_ok!(EncointerCeremonies::endorse_newcomer(
+			Origin::signed(alice.clone()),
+			cid,
+			yran.clone()
+		));
+
+		assert!(EndorseeIndex::<TestRuntime>::contains_key((cid, cindex), &yran));
+		assert!(!NewbieIndex::<TestRuntime>::contains_key((cid, cindex), &yran));
```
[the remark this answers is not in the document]
pifragile: > ok that's tested here: unregistering_newbie_works and for endorsement elsewhere

does that resolve the conversation?

ceremonies/src/lib.rs
pifragile: i dont fully get why, but i added an explicit check

pifragile: i think we should rely on continuous indices because otherwise if we dont enforece it, the assignment algorithm breaks

ceremonies/src/lib.rs
```diff
@@ -1007,6 +1014,42 @@ impl<T: Config> Pallet<T> {
...
+		<NewbieIndex<T>>::insert((cid, cindex), &last_participant, &participant_index);
+
+		<NewbieRegistry<T>>::remove((cid, cindex), &participant_count);
+		<NewbieIndex<T>>::remove((cid, cindex), &participant);
+
+		<NewbieCount<T>>::insert(
+			(cid, cindex),
+			participant_count
+				.checked_sub(1)
+				.expect("participant is registered, thus participant_count > 0; qed"),
```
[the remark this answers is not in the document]
pifragile: done. but i dont fully understand. `saturating_sub` just saturates at numeric bounds and throws no error, or am i wrong?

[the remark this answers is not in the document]
brenzi: Exactly

brenzi (approved):

ceremonies/src/lib.rs
brenzi: Yes, but even more important than ensuring functionality is to never panic.

Issue #163, opened 2022-03-07
Rename RPCs to have `encointer`-prefix


clangenb:

clangenb:
@brenzi just to not get into misunderstanding, we will go for e.g.: `ceremonies_getReputations` -> `encointer_getReputations`

And we will not go for: `encointer_ceremonyGetReputations`?

clangenb:
Closed by #203

PR #203, opened 2022-04-06
rename rpcs

pifragile:
Introduce `encointer_` prefix for RPC method names
Part of #163

### DIFF
--- a/bazaar/rpc/src/lib.rs
+++ b/bazaar/rpc/src/lib.rs
@@ -33,19 +33,19 @@ pub trait BazaarApi<BlockHash, AccountId>
 where
 	AccountId: 'static + Encode + Decode + Send + Sync,
 {
-	#[rpc(name = "encointer_bazaar_getBusinesses")]
+	#[rpc(name = "encointer_bazaarGetBusinesses")]
 	fn get_businesses(
 		&self,
 		cid: CommunityIdentifier,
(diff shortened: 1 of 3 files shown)


bazaar/rpc/src/lib.rs
```diff
@@ -33,19 +33,19 @@ pub trait BazaarApi<BlockHash, AccountId>
 where
 	AccountId: 'static + Encode + Decode + Send + Sync,
 {
-	#[rpc(name = "bazaar_getBusinesses")]
+	#[rpc(name = "encointer_bazaar_getBusinesses")]
```
clangenb, line 36: The rpc macro does not allow for double underscores, we need to do it like:

```
encointer_bazaarGetBusinsses
```

clangenb:
Now that the other PR has already been merged, you need to do the same for the aggregated account data. :)

brenzi:
please always describe a PR and mention the issue it closes

pifragile:
ok, i wasn't sure because it does not acutally close the issue (if we consider the refactoring of the rpc into a separate crate as part of the issue.) But i will mention it anyway

clangenb (approved):
LGTM, matches the result of our discussion. What is your take on this @brenzi?

brenzi:
Good to see that this does not touch the runtime at all.

PR #207, opened 2022-04-07
fix division by zero panic if phase duration is zero

brenzi:
when upgrading to full runtime from launch, we experienced a panic because phase durations were initialized to zero

### DIFF
--- a/scheduler/src/lib.rs
+++ b/scheduler/src/lib.rs
@@ -29,8 +29,8 @@ use frame_support::{
 	traits::{Get, OnTimestampSet},
 	weights::DispatchClass,
 };
-use log::info;
-use sp_runtime::traits::{CheckedAdd, CheckedDiv, One, Saturating, Zero};
+use log::{info, warn};
+use sp_runtime::traits::{CheckedDiv, One, Saturating, Zero};
 use sp_std::{ops::Rem, prelude::*};
 
 // Logger target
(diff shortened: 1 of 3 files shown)


scheduler/src/lib.rs
```diff
@@ -238,16 +242,18 @@ impl<T: Config> Pallet<T> {
 
 		let tnext = if tnext < now {
 			let gap = now - tnext;
-			let n = gap
-				.checked_div(&cycle_duration)
-				.expect("invalid phase durations: may not be zero");
-			tnext.saturating_add((cycle_duration).saturating_mul(n + T::Moment::one()))
+			if let Some(n) = gap.checked_div(&cycle_duration) {
```
clangenb, line 245: The `if let Some(...) else return Err`. Is a more verbose way of writing:

`operation.ok_or_else(|| Err...)?;`

scheduler/src/lib.rs
```diff
@@ -257,13 +263,23 @@ impl<T: Config> Pallet<T> {
...
-				.expect("overflowing timestamp");
-			Self::resync_and_set_next_phase_timestamp(next).expect("set next phase failed");
+				.saturating_add(<PhaseDurations<T>>::get(CeremonyPhaseType::REGISTERING));
+
+			if Self::resync_and_set_next_phase_timestamp(next).is_err() {
+				warn!(target: LOG, "resync ceremony phase failed");
+			};
 		} else if Self::next_phase_timestamp() < now {
-			Self::progress_phase().expect("phase progress error");
+			if Self::progress_phase().is_err() {
```
clangenb, line 280: We don't want to log the error?

clangenb (approved):
Only a beauty comment, looks good otherwise!

brenzi:
yeah, we'll fix that next time :sweat_smile:

PR #209, opened 2022-04-13
[primitives] change `CeremonyPhase` values to CamelCase

### LINKED ISSUE #208: Use CamelCase for CeremonyPhase values
This is how enums are generally defined in rust, and we do it everywhere else like this.

clangenb:
* Closes #208 

Haha, it took me ages to realize why I need to handle this enum type differently on our client side.


brenzi (approved):
Well, I see no priority in this and it will certainly cause integration work downstream. But as you wish ;-)

PR #221, opened 2022-05-25
Refactor meetup validation

pifragile:
This is a new implementation of the meetup validation. An external crate was introduced to separate all the logic from the data in order to simplify unit testing.
Also I redesigned the algorithm to be more resilient. However I found that this validation process is quite complex and probably quite hard to get right in all the details.
I tried to cover as many cases as possible, but there might still be some tricky edge cases that my algo misses.

The core piece of the new module is the algorithm that excludes participants based on their attestations and it roughly works as follows:
- Group all participants by their incoming and outgoing attestations
- Find min(min_incoming, min_outgoing)
- Check if this number is below the threshold
- If yes exclude the participants completely (ie, do not count their attestations anymore and also do not expect an attestation from them anymore)

Repeat the above steps until all participants have incoming and outgoing attestations above the threshold
(Note that the threshold decreases when participants are excluded)

This algorithm has the advantage that no subgroup smaller than half of the meetup can manipulate the meetup. 

I know that what I did is way beyond a simple refactor, but my mind saw all kind of problems arise with the old approach and it was hard to not try to fix them when having to rewrite all the code anyways.
I ask the reviewers to not only check code quality, but also the reason about the algorithm logic.


ceremonies/meetup-validation/src/tests.rs
```diff
@@ -0,0 +1,228 @@
...
+	);
+}
+
+#[test]
+fn filter_attestations_works() {
+	let participants: Vec<usize> = vec![0, 2, 4];
+	let participant_attestations: Vec<Vec<usize>> =
+		vec![vec![2], vec![], vec![0], vec![2], vec![0, 2]];
+	let expected: Vec<Vec<usize>> = vec![vec![2], vec![], vec![0], vec![2], vec![0, 2]];
+	assert_eq!(filter_attestations(&participants, &participant_attestations), expected);
```
brenzi, line 36: Strange case where input equals output. What are you testing here?

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+			if min_num_incoming_attestations < threshold_fn(participants_to_process.len()) {
+				let ps = participants_grouped_by_incoming_attestations[0].1.clone();
+				ps.clone().into_iter().for_each(|p| {
+					excluded_participants.push((p, ExclusionReason::TooFewIncomingAttestations))
+				});
+
+				// remove the participants from the included participants and the attestation vectors
+				participants_to_process.retain(|k| !ps.contains(k));
+				relevant_attestations =
+					filter_attestations(&participants_to_process, &relevant_attestations);
```
brenzi, line 106: this is redundant code with the case below

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+				.len()
+		})
+		.collect();
+	return (
+		group_participants_by_num_attestations(participants, &num_outgoing_attestations),
+		group_participants_by_num_attestations(participants, &num_incoming_attestations),
+	)
+}
+
+fn group_participants_by_num_attestations(
```
brenzi, line 192: your naming is more verbose than necessary: for this fn it doesn't matter that we're treating `participants` and `attestations`. It just takes in index and a number per index and sorts and groups. I'd prefer to just describe what the function does, not what we are using it for

ceremonies/meetup-validation/src/tests.rs
```diff
@@ -0,0 +1,228 @@
...
+
+#[test]
+fn find_majority_vote_works() {
+	let participants: Vec<usize> = vec![0, 1, 2, 3, 4];
+	let participant_votes: Vec<u32> = vec![1, 1, 2, 3, 1];
+	assert_eq!(find_majority_vote(&participants, &participant_votes), Ok((1u32, 3u32)));
+}
+
+#[test]
+fn get_updated_participants_works_case_0() {
```
brenzi, line 94: your numbered cases only differ in inputs and expected outputs. Please refactor this to a parametric test `#[rstest(...`. This will make it a lot easier to include more test cases

ceremonies/meetup-validation/src/tests.rs
```diff
@@ -0,0 +1,228 @@
...
+			attestation_threshold_fn,
+		)
+		.unwrap(),
+		updated_participants
+	);
+}
+
+#[test]
+fn get_updated_participants_works_case_1() {
+	// Scenario: One participant did not vote
```
brenzi, line 118: this can't happen by definition of `ClaimOfAttendance` type. should be "one participant voted differently"

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+pub fn get_updated_participants(
+	participants: &Vec<usize>,
+	participant_votes: &Vec<u32>,
+	participant_attestations: &Vec<Vec<usize>>,
+	attestation_threshold_fn: fn(usize) -> usize,
+) -> Result<UpdatedParticipants, MeetupValidationError> {
+	let mut updated_participants =
+		UpdatedParticipants { included: participants.clone(), excluded: vec![] };
+	updated_participants.exclude_participants(get_excluded_participants_no_vote(
```
brenzi, line 14: why is no_vote different than wrong_vote? I don't think we need special treatment. Especially because the `ClaimOfAttendance` type always contains a vote. So, if there is no vote, the only cause is that the participant did not receive any attestations from others and will be excluded anyway.
I think what you want to achieve is that the winning vote isn't zero of an assignment of 10 participants has 4 legit  attendees voting 4, right? If that is the case, then please add some comment or make naming clearer. Currently the logic is difficult to read

ceremonies/meetup-validation/src/tests.rs
```diff
@@ -0,0 +1,228 @@
...
+		)
+		.unwrap(),
+		updated_participants
+	);
+}
+
+#[test]
+fn get_updated_participants_works_case_3() {
+	// Scenario: 5 participants
+	// 0 has a broken phone and can only be attested by 1 and 2 and attests nobody
```
brenzi, line 171: what exactly is wrong with this phone? camera not working? I think you're combining two problems here: `0` doesn't attest anyone and is not attested by some. I'd separate these cases

ceremonies/meetup-validation/src/tests.rs
```diff
@@ -0,0 +1,228 @@
...
+		.unwrap(),
+		updated_participants
+	);
+}
+
+#[test]
+fn get_updated_participants_works_case_4() {
+	// Scenario: 5 participants
+	// 0 has a broken screen and can only be attested by 1 and 2 but attests everybody else
+	// we expect that everybody except for 0 gets a reward
```
brenzi, line 203: so here we have the actual improvement in behaviour over the status quo: This mitigates the problem that someone with ill intent (or good intent but bad gear) can harm others

ceremonies/meetup-validation/src/tests.rs
brenzi: and I'd like to see many more cases:
* assignment size from 3..18
* attendance size: n-0, n-1.....down to zero
* attestment size n-1 down to zero
* all real cases we had so far

brenzi (changes requested):
I like this PR a lot. 
Still: there aren't enough tests to convince me that the algo does what it should. In principle, I understand the algo and I think it really does enforce a fully connected, fully reciprocal adjacency matrix with the current settings. So I believe the algo is correct. But I need better test coverage of different cases

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+			&relevant_attestations,
+		);
+
+		let min_num_outgoing_attestations = participants_grouped_by_outgoing_attestations[0].0;
+		let min_num_incoming_attestations = participants_grouped_by_incoming_attestations[0].0;
+
+		if min_num_incoming_attestations < min_num_outgoing_attestations {
+			if min_num_incoming_attestations < threshold_fn(participants_to_process.len()) {
+				let ps = participants_grouped_by_incoming_attestations[0].1.clone();
+				ps.clone().into_iter().for_each(|p| {
```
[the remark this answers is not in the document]
pifragile: if i remove either of the clones the code wont compile anymore. maybe i make some fundamental thinking mistakes about my types that makes this cloning necessary🤷‍♂️ i am happy to learn :)

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl UpdatedParticipants {
+	pub fn exclude_participants(&mut self, excluded: Vec<(usize, ExclusionReason)>) {
+		self.included = self
+			.included
+			.clone()
+			.into_iter()
+			.filter(|&i| !excluded.iter().any(|p| p.0 == i))
+			.collect();
```
clangenb, line 245:
```suggestion
                  self.included.retain(|&i| !excluded.iter().any(|p| p.0 == i));

```

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+	Ok((n_confirmed, vote_count))
+}
+
+fn filter_attestations(
+	participants: &Vec<usize>,
+	participant_attestations: &Vec<Vec<usize>>,
+) -> Vec<Vec<usize>> {
+	// filter out participants from the attestation vectors that are not in the participants vector anymore.
+	participant_attestations
+		.clone()
```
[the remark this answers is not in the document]
clangenb: It is not good practice to pass references to a function when you need to call clone inside the functions. Better pass an owned value, this potentially removes the amount because the value might not be needed later on the outside. You can also be more ergonomic here:

```
fn filter_attestations(
	participants: &Vec<usize>,
	participant_attestations: Vec<Vec<usize>>,
) -> Vec<Vec<usize>> {
	// filter out participants from the attestation vectors that are not in the participants vector anymore.
	participant_attestations
		.into_iter()
		.map(|a| a.into_iter().filter(|j| participants.contains(j)).collect())
		.collect()
}
```

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+				ps.clone().into_iter().for_each(|p| {
+					excluded_participants.push((p, ExclusionReason::TooFewOutgoingAttestations))
+				});
+
+				// remove the participants from the included participants and the attestation vectors
+				participants_to_process.retain(|k| !ps.contains(k));
+				relevant_attestations =
+					filter_attestations(&participants_to_process, &relevant_attestations);
+				continue
+			}
```
clangenb, line 121: The if and else statement here are identical except for the exclusion reason, you should either define a function or a closure to remove the redundant definition

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+	for i in participants {
+		match participant_votes[*i] {
+			v if v > 0 => continue,
+			_ => excluded_participants.push(*i),
+		}
+	}
+	excluded_participants
+		.into_iter()
+		.map(|p| (p, ExclusionReason::NoVote))
+		.collect()
```
clangenb, line 51: `collect` is O(n) so better try to avoid that:

```suggestion
	       let mut excluded_participants: Vec<(usize, ExclusionReason)> = vec![];
	       for i in participants {
		match participant_votes[*i] {
			v if v > 0 => continue,
			_ => excluded_participants.push((*i, ExclusionReason::NoVote)),
		}
	}
       excluded_participants
```

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+) -> Vec<(usize, ExclusionReason)> {
+	let mut excluded_participants: Vec<usize> = vec![];
+	for i in participants {
+		if participant_votes[*i] != n_confirmed {
+			excluded_participants.push(*i)
+		}
+	}
+	excluded_participants
+		.into_iter()
+		.map(|p| (p, ExclusionReason::WrongVote))
```
clangenb, line 67: Same as above

clangenb (commented):
Looks promising, this feedback is only intermediary.

Further, it makes it very hard to read the code with all the bare `(usize, Vec<usize>)` I would love to see some type aliases to make the code more readable, or at least write in the function doc, what the values stand for.

ceremonies/meetup-validation/src/lib.rs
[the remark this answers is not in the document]
clangenb: Edit: Even more concise and efficient:

```
	participant_attestations
		.into_iter()
		.map(|mut a| {
			a.retain(|j| participants.contains(j));
			a
		})
		.collect()
```

ceremonies/meetup-validation/src/tests.rs
pifragile: about using rstest i have some concerns. i built it now and i just want to give you a feeling of how it looks:
```
#[rstest(participants, participant_votes, participant_attestations, participant_judgements,
// Scenario: Everyone attests everyone
case(vec![0, 1, 2],
	vec![3, 3, 3],
	vec![vec![1, 2], vec![0, 2], vec![0, 1]],
	ParticipantJudgements { legit: participants.clone(), excluded: vec![] }
),
// Scenario: One participant did not vote
case(vec![0, 1, 2, 3],
	vec![0, 3, 3, 3],
	vec![vec![1, 2, 3],vec![0, 2, 3], vec![0, 1, 3], vec![0, 1, 2]],
	ParticipantJudgements {
		legit: vec![1, 2, 3],
		excluded: vec![ExcludedParticipant { index: 0, reason: ExclusionReason::NoVote }],
	}
),
// Scenario: One participant did not vote like majority
case(vec![0, 1, 2, 3],
	vec![1, 3, 3, 3],
	vec![vec![1, 2, 3], vec![0, 2, 3], vec![0, 1, 3], vec![0, 1, 2]],
	ParticipantJudgements {
		legit: vec![1, 2, 3],
		excluded: vec![ExcludedParticipant { index: 0, reason: ExclusionReason::WrongVote }],
	}
),
// Scenario: 5 participants
// 0 has a broken phone and can only be attested by 1 and 2 and attests nobody
// we expect that everybody except for 0 gets a reward
case(vec![0, 1, 2, 3, 4],
vec![5, 5, 5, 5, 5],
	vec![vec![], vec![0, 2, 3, 4], vec![0, 1, 3, 4], vec![1, 2, 4], vec![1, 2, 3]],
	ParticipantJudgements {
		legit: vec![1, 2, 3, 4],
		excluded: vec![ExcludedParticipant { index: 0, reason: ExclusionReason::TooFewOutgoingAttestations }],
	}
),
// Scenario: 5 participants
// 0 has a broken phone and can only be attested by 1 and 2 and attests nobody
// we expect that everybody except for 0 gets a reward
case(vec![0, 1, 2, 3, 4],
	vec![5, 5, 5, 5, 5],
	vec![vec![1, 2, 3, 4], vec![0, 2, 3, 4], vec![0, 1, 3, 4], vec![1, 2, 4], vec![1, 2, 3]],
	ParticipantJudgements {
		legit: vec![1, 2, 3, 4],
		excluded: vec![ExcludedParticipant { index: 0, reason: ExclusionReason::TooFewIncomingAttestations }],
	}
)
)]
```

considering that we do not have code highlighting and auto format in the rstest code, i am a bit worried readbility and maintainability. @clangenb what do you think? do you know of any plugins that give us code highlighting and auto format for rstest?

clangenb: No, I don't, but formatting should be supported in attribute type proc macros.

Also, I would recommend to give your test cases names (not only documentation) with: `case::scenario5`

Maybe coud abuse the `fixures` attribute to get something like:

```
case::scenario1(test_cases(1))
fn my_test(test_case: TestCase) {...}
```

where the `test_cases(i: usize)` is a fixture that simply returns the ith element or our test cases.

ceremonies/src/lib.rs
```diff
@@ -1651,6 +1623,68 @@ impl<T: Config> Pallet<T> {
...
+			) {
+				Some(attestees) => attestees,
+				None => vec![],
+			};
+			// convert AccountId to local index
+			let attestation_indices = attestations
+				.into_iter()
+				.map(|p| meetup_participants.iter().position(|s| *s == p))
+				.filter(|i| i.is_some())
+				.map(|i| i.unwrap())
```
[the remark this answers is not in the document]
pifragile: how can i simplify it?

ceremonies/meetup-validation/src/tests.rs
pifragile: this would be possible, but also does not look very handy, because you would have to manually add a `#case` for every testcase. i also looked at `#values` but i guess it does not work very well because cannot pass a vec of testcases in there. so after some thought i came to the conclusion that this fee0634780b1888b7f405b4f1d6ef1e6cf4b468b might be the best solution. what do you think?

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+			excluded_participants.push((*i, ExclusionReason::WrongVote))
+		}
+	}
+	excluded_participants
+}
+
+/// The algorithm works as follows:
+/// We find the participant with the fewest incoming/outgoing attestations
+/// If it if below the threhsold, we exclude the participant
+fn get_excluded_participants_num_attestations(
```
clangenb, line 79: I am not so happy with this function I think it does too much, why do we not separate the handling of the exclusion based on incoming and outgoing attestations?

I think it should be possible to do separate them without getting a performance overhead.

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+	participants: &Participants,
+	participant_attestations: &Attestations,
+	threshold_fn: fn(usize) -> usize,
+) -> Vec<(usize, ExclusionReason)> {
+	let mut relevant_attestations =
+		filter_attestations(participants, participant_attestations.clone());
```
clangenb, line 85: again, passing a reference to clone inside.

clangenb (commented):
Some more intermediary feedback

ceremonies/meetup-validation/src/lib.rs
pifragile: the algorithm works by selecting always the participant with the fewest incoming OR outgoing attestations and removing it from the graph. so participant 1 could have 0 incoming attestations an gets removed. then participant 5 could have only 2 outgoing attestations and be removed. then participant 7 could have only 3 incoming and be removed. so we cannot separate the two unfortunatley.

ceremonies/src/lib.rs
```diff
@@ -383,11 +384,98 @@ pub mod pallet {
...
+			};
+			participants_eligible_for_rewards = participant_judgements.legit;
+			// emit events
+			participant_judgements.excluded.iter().for_each(|p| {
+				let participant = meetup_participants[p.index].clone();
+				let event = match p.reason {
+					ExclusionReason::NoVote => {
+						debug!(target: LOG, "participant {:?} received no rewards because they did not vote. meetup_index: {:?}, cid: {:?} cindex: {:?}", participant, meetup_index, cid, cindex);
+						Event::NoRewardBecauseDidNotVote
+					},
```
clangenb, line 455: I think we don't need to do this match statement here (see comment below). Then we can also just use one debug statement, with the same text, we print the specific reason anyhow.

ceremonies/meetup-validation/src/tests.rs
brenzi: I think it is very important to have a good dev experience here and it may not be overly complicated to read: So I suggest the following:

1. use rstest to vary two parameters only: number of assignees and number attendees (optionally, add a scenario-specific parameter 
2. assume fully connected and fully reciprocal adjacency matrix and then - based on the scenario - remove attestations selectively. Your current solution is way too verbose and error-prone because it's impossible to read and grasp a 10x10 adjacency matrix IMO  
3. write one rstest per scenario ("n_attendees_do_not_attest_anyone", "attendee_is_not_attested_by_n_others")

ceremonies/src/lib.rs
```diff
@@ -389,11 +390,93 @@ pub mod pallet {
...
+				.ok_or(<Error<T>>::GetMeetupParticipantsError)?;
+			let (participant_votes, participant_attestations) =
+				Self::gather_meetup_validation_data(cid, cindex, meetup_participants.clone());
+
+			// initialize an array of local participant indices that are eligible for the reward
+			// indices will be deleted in the following based on various rules
+			let mut participants_eligible_for_rewards: Vec<usize> =
+				(0..meetup_participants.len()).collect();
+
+			let attestation_threshold_fn = |i: usize| max(i.saturating_sub(1), 1);
```
clangenb, line 425: I guess the right-hand side max should be 3. This is what we decided is the minimum meetup size for a successful meetup.

ceremonies/src/lib.rs
```diff
@@ -1659,6 +1603,66 @@ impl<T: Config> Pallet<T> {
...
+				Some(attestees) => attestees,
+				None => vec![],
+			};
+			// convert AccountId to local index
+			let attestation_indices = attestations
+				.into_iter()
+				.filter_map(|p| meetup_participants.iter().position(|s| *s == p))
+				.collect();
+
+			participant_attestations.push(attestation_indices);
```
clangenb, line 1660: We can skip the collect here:

```
		let attestation_indices = attestations
				.into_iter()
				.filter_map(|p| meetup_participants.iter().position(|s| *s == p));

		participant_attestations.extend(attestation_indices);
```

ceremonies/src/lib.rs
```diff
@@ -1659,6 +1603,66 @@ impl<T: Config> Pallet<T> {
...
+	fn issue_rewards(
+		cid: CommunityIdentifier,
+		cindex: CeremonyIndexType,
+		meetup_idx: MeetupIndexType,
+		meetup_participants: Vec<T::AccountId>,
+		participants_indices: Vec<usize>,
+	) {
+		let reward = Self::nominal_income(&cid);
+		for i in &participants_indices {
+			let participant = &meetup_participants[*i];
```
clangenb, line 1615: Oh, we have another raw index operation here.

ceremonies/meetup-validation/src/meetup_scenario_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+	case(3, 3, 2)
+)]
+fn attendee_is_not_attested_by_n_others(meetup_size: usize, num_attendees: usize, n: usize) {
+	let participants: Participants = (0..meetup_size).collect();
+	let mut participant_votes = vec![num_attendees as u32; num_attendees];
+	participant_votes.append(&mut vec![0u32; meetup_size - num_attendees]);
+	let mut participant_attestations: Attestations = vec![];
+
+	for i in 0..num_attendees {
+		let first_index = if i <= n { 1 } else { 0 };
```
clangenb, line 88: I can't understand the significance of the first index here.

clangenb (commented):
There has been some serious work in this PR. Still some comments, but I like it!

ceremonies/meetup-validation/src/meetup_scenario_tests.rs
pifragile: participant with index 0 is the guy that is not attested by n others. so everyone with index <= n will not attest the participant 0 and will therefore attest everyone starting at index 1

ceremonies/src/lib.rs
pifragile: @brenzi decided to leave it as is for the moment being

ceremonies/src/lib.rs
pifragile: `participant_attestations: Vec<Vec<usize>>`
so i want to push a vec to a vec and not extend it

clangenb: Ok, I see. By bad. :)

clangenb:
Alright, massive PR! And I think the iterative review was worth it! LGTM!

clangenb (approved):

brenzi (approved):
OK. let's move ahead.
I don't think this is the final word on the testing, but it is such a great improvement overall that I don't want to delay this PR any more

thanks for the good ideas work!

PR #222, opened 2022-05-25
Mark issuance as completed, when evaluating a failed meetup

clangenb:


brenzi (approved):

clangenb:
Now that I think about it, how do clients learn whether a meetup has failed or not? Maybe we should change the `IssueRewards` storage to contain an enum:  `IssuanceResult`

```
pub enum IssuanceResult {
    Ok,
    Failed,
}
```

brenzi:
Sounds good and necessary

brenzi:
@clangenb will you implement that change or do you want to hand it over?

clangenb:
I don't really care, I thought I will wait until the refactoring PR is finished. but I only wait if both can land in the next runtime upgrade

pifragile:
@brenzi we might need a storage migration for this. this sounds somewhat tedious and is badly documented... should we still do it? could take some time, benefit is not huge, as we could also implement the change as @clangenb initially suggested without the new type. on the other hand we could gather knowledge on how to do these migrations.. what do you think?

brenzi:
Sooner or later we'll have to do a storage migration, but I'd rather prefer later ;-)
I guess for now we just mark the meetup complete and one needs to check events to learn more details. So I guess that was the first commit here, right?

PR #223, opened 2022-05-25
implement transfer_all

pifragile:
I implemented it such that we take the full available balance.
[Here](https://docs.substrate.io/v3/runtime/weights-and-fees/) I read that

"You should note that transaction fees are withdrawn before the transaction is executed. After the transaction is executed, the transaction weight can be adjusted to reflect the actual resources the transaction used. If a transaction uses fewer resources than expected, the transaction fee is corrected and the adjusted transaction fee is deposited."

So i guess it is ok to take the full balance as the fees were already deducted. What I couldn't test was what happens if some fees are returned after the txn and the account was deleted.


brenzi:
I suggest to extend the unit test to include demurrage: just set a higher blocknumber after issue and see if the logic can cope

brenzi:
And also, test that the account has been removed

Moreover: with this test fixture, fees are out of scope. Maybe we should add this to integration testing (bootstrap-demo-community)?

pifragile:
> And also, test that the account has been removed

The removal of the account is tested [here](https://github.com/encointer/pallets/pull/223/files#diff-46f94b84fb4cacb8541882fdd1903b72ee421c03cda72a0e479c60f23bc4ed39R291)


> Moreover: with this test fixture, fees are out of scope. Maybe we should add this to integration testing (bootstrap-demo-community)?

Can we pay fees in CC in the integration tests?

clangenb:
> Can we pay fees in CC in the integration tests?

The basis has been implemented in the  api-client here: https://github.com/scs/substrate-api-client/pull/232

But we need to update the encointer-client to support this.

brenzi:
@pifragile could you implement a new cli comman, wrap it in python and include it in our CI script bootstrap-demo-community plz? As @clangenb suggested above

brenzi:
Changed my mind: Let's merge this and fix if needed

balances/src/lib.rs
```diff
@@ -260,6 +273,13 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// Remove an account from a community
+	fn remove_account(cid: CommunityIdentifier, who: &T::AccountId) -> DispatchResult {
+		<Balance<T>>::remove(cid, who);
```
brenzi, line 278: this is a critical, destructive function. please ensure that the balance is below ED before removing the account 

even if the current usage ensures that, we should not assume that devs know what they're doing ;-)

brenzi (changes requested):

balances/src/lib.rs
pifragile: haha ok done

balances/src/lib.rs
```diff
@@ -111,6 +111,19 @@ pub mod pallet {
...
+		#[pallet::weight((<T as Config>::WeightInfo::transfer_all(), DispatchClass::Normal))]
+		pub fn transfer_all(
+			origin: OriginFor<T>,
+			dest: T::AccountId,
+			cid: CommunityIdentifier,
+		) -> DispatchResultWithPostInfo {
+			let from = ensure_signed(origin)?;
+			let amount = Self::balance(cid, &from);
+			Self::do_transfer(cid, from.clone(), dest, amount)?;
+			Self::remove_account(cid, &from)?;
```
clangenb, line 124: This should be incorporated into `do_transfer` we also want to reap the account if the balance falls below the ED in normal transfers.

clangenb (commented):
Looks nice and clean, I only have one remark!

balances/src/lib.rs
pifragile: should it be in `do_transfer` or maybe rather in `balance_entry_updated` to make sure that this mechanism also kicks in when demurrage puts an account below ED? and generally, should this be part of this PR?

clangenb: I think in `do_transfer` suffices for now because otherwise the demurrage is only applied when you get money. So the reaping is most likely not kicking in anyhow, and it might be a bit more tricky there to not make any errors.

pifragile: ok done

balances/src/lib.rs
pifragile: uuh good catch, i forgot to remove the call to `remove_account` in `transfer_all`

pifragile: also added a check

brenzi (approved):

PR #231, opened 2022-06-15
new attesting flow

### LINKED ISSUE #215: simplify submission of attestations. no need for signed claimOfAttendance
In my opinion, the design around signed claims of attendances is not very intuitive.
I don't see what benefit we get from the fact that the claims have a signature. Does it add any security?
Also the claims hold a lot of redundant information that needs to be processed on chain.

I would propose a new design, where meetup participants just send the following info after a meetup:
```
ceremony_index,
community_identifier,
meetup_index,
location,
timestamp,
list of public keys of meetup participants // instead of sigend attestations
```

Then, we would read the vote on the number of participants from the length of the list of public keys rather than from the claims.
This would simplify things and make it more intuitive. I cannot see any security issues with this new design. Am I missing something?

What are your thoughts on this @brenzi @clangenb?

pifragile:
* Closes #215


ceremonies/src/lib.rs
```diff
@@ -1669,6 +1731,79 @@ impl<T: Config> Pallet<T> {
...
+
+		Ok((cindex, meetup_index, meetup_participants, meetup_location, meetup_time))
+	}
+
+	fn add_attestations_to_registry(
+		participant: T::AccountId,
+		cid: &CommunityIdentifier,
+		cindex: CeremonyIndexType,
+		meetup_index: MeetupIndexType,
+		meetup_participants: &Vec<T::AccountId>,
```
clangenb, line 1762: I have seen this for a while now. Instead of `&Vec<T::Account>` it is the better practice to put `&[T::Account]` because `&Vec<T::AccountId>` can always be coerced to `&[T::Account]`, but not vice versa.

ceremonies/src/lib.rs
```diff
@@ -170,6 +171,63 @@ pub mod pallet {
 			Ok(().into())
 		}
 
+		#[pallet::weight(10_000)]
+		pub fn attest_attendees(
+			origin: OriginFor<T>,
+			cid: CommunityIdentifier,
+			number_of_participants_vote: u32,
+			timestamp: T::Moment,
```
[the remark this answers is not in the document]
clangenb: I have just discussed with @pifragile. If we really want to use the time to ensure that the meetup has been within the meetup time toleration, we should do something with the timestamp pallet. So we can check upon arrival of an extrinsics that the current timestamp is within a reasonable range.

ceremonies/src/benchmarking.rs
```diff
@@ -257,6 +257,35 @@ benchmarks! {
...
+		let attestor_account = account_id::<T>(&attestor);
+
+		assert_ok!(Pallet::<T>::register_participant(
+			RawOrigin::Signed(attestor_account.clone()).into(),
+			cid,
+			Some(fake_last_attendance_and_get_proof::<T>(&attestor, cid)))
+		);
+
+		let attestees =  register_users::<T>(cid, 2, 7).into_iter().map(|u| account_id::<T>(&u)).collect();
+
```
clangenb, line 273: To ensure that we test the worst-case scenario, we should increase the meetup size a bit. Maybe up to 15?

primitives/src/common.rs
```diff
@@ -80,3 +82,27 @@ pub enum IpfsValidationError {
 	InvalidLength(u8),
 	InvalidBase58(Bs58Error),
 }
+
+pub trait AbsDiff<T> {
+	fn abs_diff(&self, other: T) -> T;
+}
```
clangenb, line 88: I think this is better in `primitives/core`

primitives/src/common.rs
```diff
@@ -80,3 +82,27 @@ pub enum IpfsValidationError {
...
+
+impl<T: BaseArithmetic + Copy> AbsDiff<T> for T {
+	fn abs_diff(&self, other: T) -> T {
+		if *self > other {
+			return *self - other
+		} else {
+			return other - *self
+		}
+	}
+}
```
clangenb, line 98: So nice!

clangenb (commented):
Some minor comments, but it looks very nice and clean already!

ceremonies/src/lib.rs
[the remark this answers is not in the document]
brenzi: But an extrinsic may be sent a lot later if people don't have internet at the meetup place

I agree that there is no additional security though as we trust the client for providing an honest tine anyway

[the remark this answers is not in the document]
brenzi: Dropping the timestamp arg is fine

ceremonies/src/benchmarking.rs
pifragile: i tried, but then the participants are split over 2 meetups by the assignment algo

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -1,6 +1,8 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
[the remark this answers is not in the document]
pifragile: 😅

ceremonies/src/lib.rs
```diff
@@ -170,6 +170,57 @@ pub mod pallet {
 			Ok(().into())
 		}
 
+		#[pallet::weight(10_000)]
```
clangenb, line 173: Ah, we need to add the benchmark here.

clangenb (commented):
Minor comment looks good otherwise!

Issue #178, opened 2022-03-12
[rpc] verify meetup already during attesting phase


brenzi:
Assuming that all participants are online and send their attestations right after the ceremony meetup, we could offer a validation function that actually performs a full validation of the meetup right after it happened. This way, you immediately know if you will get your income or not

clangenb:
* Closed by #233, which does even allow early payout.

PR #233, opened 2022-06-17
early rewards

pifragile:
* Coses #178

### DIFF
--- a/ceremonies/meetup-validation/src/lib.rs
+++ b/ceremonies/meetup-validation/src/lib.rs
@@ -22,14 +22,17 @@ pub fn get_participant_judgements(
 	participant_attestations: &Attestations,
 	attestation_threshold_fn: fn(usize) -> usize,
 ) -> Result<ParticipantJudgements, MeetupValidationError> {
-	let mut participant_judgements =
-		ParticipantJudgements { legit: participants.clone(), excluded: vec![] };
+	let mut participant_judgements = ParticipantJudgements {
+		legit: participants.clone(),
+		excluded: vec![],
+		early_rewards_possible: false,
+	};
 	participant_judgements.exclude_participants(get_excluded_participants_no_vote(
 		&participant_judgements.legit,
 		participant_votes,
 	)?);
 
-	let (n_confirmed, _num_votes) =
+	let (n_confirmed, _num_votes, vote_is_unanimous) =
 		find_majority_vote(&participant_judgements.legit, participant_votes)?;
 
 	participant_judgements.exclude_participants(get_excluded_participants_wrong_vote(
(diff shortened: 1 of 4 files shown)


clangenb (commented):
Cool, this looks very good! I only have two details!

clangenb (approved):
Looks good to me!

ceremonies/meetup-validation/src/lib.rs
```diff
@@ -38,15 +41,83 @@ pub fn get_participant_judgements(
...
+		.enumerate()
+		.map(|(i, v)| !(legit_participants.contains(&i)) || v.len() == (n_confirmed - 1) as usize)
+		.all(|item| item)
+}
+
+fn attestation_graph_is_fully_connected(
+	legit_participants: Participants,
+	participant_attestations: Attestations,
+) -> bool {
+	for (i, mut attestations) in participant_attestations.into_iter().enumerate() {
```
brenzi, line 83: what happens, if `i` attested people that fell out of the legit participant set? I believe your logic with sorting and testing equality would not see this as fully connected. but it should

Anyway, sorting twice within a for loop doesn't strike me as efficient. Wouldn't a nested for loop through legit participants be more efficient?

ceremonies/src/lib.rs
```diff
@@ -443,11 +443,17 @@ pub mod pallet {
...
-			if <encointer_scheduler::Pallet<T>>::current_phase() != CeremonyPhaseType::Registering {
+			let current_phase = <encointer_scheduler::Pallet<T>>::current_phase();
+			if ![CeremonyPhaseType::Registering, CeremonyPhaseType::Attesting]
+				.contains(&current_phase)
+			{
 				return Err(<Error<T>>::WrongPhaseForClaimingRewards.into())
 			}
 
-			let cindex = <encointer_scheduler::Pallet<T>>::current_ceremony_index() - 1; //safe; cindex comes from within
+			let cindex_offset = if current_phase == CeremonyPhaseType::Registering { 1 } else { 0 };
```
brenzi, line 453: Detail: what about a match statement on the current_phase which solves the ensure Phase and the offset at the same time? might be more readable?

brenzi (commented):
I'll need to look at the tests in detail.

ceremonies/meetup-validation/src/lib.rs
pifragile: this test is made before any participants can fall out of the set, because the attestation graph needs to be fully connected for all participants in order for early payout to be possible, right?

about the efficiency. my implementation:
O(n) (loop over attestations) * O(n^2) (sorting, worstcase)

nested loops:
O(n) (loop over attestations) * O(n) (loop over each elem in attestations) * O(n) (check if legit_participant contains elem)

so in asymptotic complexity they are the same.
i agree that in the first implementation we have some overhead because it is 2*O(n^2), but i would argue that for those small input sizes it shouldnt really matter.

but you decide, should i change it to the nested loop?

brenzi: nah, its bounded, so I guess it's fine. Do we already have a worst case benchmark for this claim function? that will tell us the price

brenzi:
I need more time to review the tests, do not merge yet

brenzi (approved):
actually, I prefer to review the big picture once all PR's are done, so I trust @clangenb 's review for now

PR #234, opened 2022-06-18
add `Issued` event per account

brenzi:
this is necessary to reconstruct entire balance history for an account based on events only

more robust testing of event deposit: Do not assume a specific order of events, just check if an event has been deposited


clangenb:
I don't think it should be a problem that the event is from another pallet, you can just import the specific event enum:

`use encointer_balances::Event as BalancesEvent;`

brenzi:
cheers, done. and a bit more...

test-utils/src/helpers.rs
```diff
@@ -98,6 +98,17 @@ pub fn event_at_index<T: frame_system::Config>(index: usize) -> Option<T::Event>
...
+pub fn event_deposited<T: frame_system::Config>(desired_event: T::Event) -> bool {
+	let events = frame_system::Pallet::<T>::events();
+	for eventrec in events.iter() {
+		let frame_system::EventRecord { event, .. } = eventrec;
+		if *event == desired_event {
+			return true
+		}
+	}
+	return false
+}
```
clangenb, line 110: Nice utility!

balances/src/lib.rs
```diff
@@ -164,6 +164,8 @@ pub mod pallet {
 		Endowed { cid: CommunityIdentifier, who: T::AccountId, balance: BalanceType },
 		/// Token transfer success `[community_id, from, to, amount]`
 		Transferred(CommunityIdentifier, T::AccountId, T::AccountId, BalanceType),
+		/// Token issuance success `[community_id, beneficiary, amount]`
+		Issued(CommunityIdentifier, T::AccountId, BalanceType),
```
clangenb, line 168: In the future, I'd like to have event enums named fields, it's much clearer. But this should be a separate task.

clangenb (approved):
Looks good to me.

PR #241, opened 2022-06-24
[ceremonies] fix purging of a community and community ceremonies

clangenb:
The purging was untested, and was not done cleanly. This PR fixes the purging of community/-ceremonies and adds a test.

Fixes:
* We essentially had a storage leak: There were residues in the map when we purged the community ceremony before because the default value was set in the clean-up instead of removing the key. This was correct at some point as far as a remember, but it was bricked after substrate introduced the `OptionQuery` and `ValueQuery` parameters. As we did not test this, we did not notice the change in behaviour.

This PR fixes the above and tests purging the community ceremony.

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -1446,9 +1446,11 @@ impl<T: Config> Pallet<T> {
 	fn purge_community(cid: CommunityIdentifier) {
 		let current = <encointer_scheduler::Pallet<T>>::current_ceremony_index();
 		let reputation_lifetime = Self::reputation_lifetime();
-		for cindex in max(current - reputation_lifetime, 0)..current {
+
+		for cindex in max(current.saturating_sub(reputation_lifetime), 0)..current {
 			if cindex > reputation_lifetime {
-				Self::purge_registry(cindex - reputation_lifetime - 1);
+				let purge_index = current.saturating_sub(reputation_lifetime).saturating_sub(1);
+				Self::purge_registry(purge_index);
 			}
 		}
 		<encointer_communities::Pallet<T>>::remove_community(cid);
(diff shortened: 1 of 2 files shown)


ceremonies/src/lib.rs
```diff
@@ -1446,9 +1446,11 @@ impl<T: Config> Pallet<T> {
 	fn purge_community(cid: CommunityIdentifier) {
 		let current = <encointer_scheduler::Pallet<T>>::current_ceremony_index();
 		let reputation_lifetime = Self::reputation_lifetime();
-		for cindex in max(current - reputation_lifetime, 0)..current {
+
+		for cindex in max(current.saturating_sub(reputation_lifetime), 0)..current {
```
pifragile, line 1450: `max` can be removed as we can never go under 0

ceremonies/src/lib.rs
```diff
@@ -1446,9 +1446,11 @@ impl<T: Config> Pallet<T> {
...
 			if cindex > reputation_lifetime {
-				Self::purge_registry(cindex - reputation_lifetime - 1);
+				let purge_index = current.saturating_sub(reputation_lifetime).saturating_sub(1);
```
pifragile, line 1452: should be `cindex` instead of `current`

ceremonies/src/lib.rs
```diff
@@ -1837,12 +1839,14 @@ impl<T: Config> OnCeremonyPhaseChange for Pallet<T> {
 			CeremonyPhaseType::Attesting => {},
 			CeremonyPhaseType::Registering => {
 				let cindex = <encointer_scheduler::Pallet<T>>::current_ceremony_index();
+				let last_index = cindex.saturating_sub(1);
```
pifragile, line 1842: all changes from here onward can be reverted, because they will be fixed in #243

ceremonies/src/lib.rs
```diff
@@ -1248,33 +1248,33 @@ impl<T: Config> Pallet<T> {
 
 		<BootstrapperRegistry<T>>::remove_prefix(cc, None);
 		<BootstrapperIndex<T>>::remove_prefix(cc, None);
-		<BootstrapperCount<T>>::insert(cc, 0);
+		<BootstrapperCount<T>>::remove(cc);
```
clangenb, line 1252: Actually, remove the storage keys instead of setting them to the default value. The implementation from before leaves us unfortunately some storage residues...

ceremonies/src/lib.rs
clangenb: Thanks, this review made me actually realize that the function before was not even correct. :)

ceremonies/src/lib.rs
```diff
@@ -1446,11 +1446,13 @@ impl<T: Config> Pallet<T> {
...
 		let current = <encointer_scheduler::Pallet<T>>::current_ceremony_index();
 		let reputation_lifetime = Self::reputation_lifetime();
-		for cindex in max(current - reputation_lifetime, 0)..current {
-			if cindex > reputation_lifetime {
-				Self::purge_registry(cindex - reputation_lifetime - 1);
-			}
+
+		for cindex in current.saturating_sub(reputation_lifetime)..=current {
+			Self::purge_community_ceremony_internal((cid, cindex));
 		}
```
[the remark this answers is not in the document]
clangenb: The above statement is wrong. We fail to clean up the registry because we only iterate through existing communities, which caused #246.

brenzi (approved):
LGTM

PR #243, opened 2022-07-12
fix purging of inactive communities

pifragile:
* Closes https://github.com/encointer/encointer-parachain/issues/138

The issue was simple: The check if cc was issued in cindex-1 was performed at the very beginning of the registration phase. The problem is that the claiming of cc for cindex-1 only happens in this registration phase after the check.

Fixed by checking for cindex-2.
Added tests and events.
Also renamed `get_inactive_communities_works` -> `update_inactivity_counters_works` beacuse `get` suggests side-effect-freeness which is not the case here

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -1838,6 +1838,15 @@ impl<T: Config> OnCeremonyPhaseChange for Pallet<T> {
 	fn on_ceremony_phase_change(new_phase: CeremonyPhaseType) {
 		match new_phase {
 			CeremonyPhaseType::Assigning => {
+				let inactives = Self::update_inactivity_counters(
+					<encointer_scheduler::Pallet<T>>::current_ceremony_index().saturating_sub(1),
+					Self::inactivity_timeout(),
+					<encointer_communities::Pallet<T>>::community_identifiers(),
+				);
+				for inactive in inactives {
+					Self::purge_community(inactive);
+				}
+
 				Self::generate_all_meetup_assignment_params();
 			},
 			CeremonyPhaseType::Attesting => {},
@@ -1848,16 +1857,6 @@ impl<T: Config> OnCeremonyPhaseChange for Pallet<T> {
 					Self::purge_registry(
 						cindex.saturating_sub(Self::reputation_lifetime()).saturating_sub(1),
 					);
-				}
-				let inactives = Self::update_inactivity_counters(
-					// check inactivity with delay, such that rewards can be claimed before the inactivity check happens
-					<encointer_scheduler::Pallet<T>>::current_ceremony_index().saturating_sub(2),
-					// compensate for the delay
-					Self::inactivity_timeout().saturating_add(1),
-					<encointer_communities::Pallet<T>>::community_identifiers(),
-				);
-				for inactive in inactives {
-					Self::purge_community(inactive);
 				}
 			},
 		}
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -689,6 +689,9 @@ pub mod pallet {
 			account: T::AccountId,
 			reason: ExclusionReason,
 		},
+
+		/// The inactivity counter of a community has been increased
+		InactivityCounterUpdated(CommunityIdentifier, u32),
 	}
 
 	#[pallet::error]
@@ -1422,7 +1425,7 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
-	fn get_inactive_communities(
+	fn update_inactivity_counters(
 		cindex: u32,
 		inactivity_timeout: u32,
 		cids: Vec<CommunityIdentifier>,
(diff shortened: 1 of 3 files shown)


ceremonies/src/lib.rs
```diff
@@ -1431,12 +1434,15 @@ impl<T: Config> Pallet<T> {
 		for cid in cids {
 			if <IssuedRewards<T>>::iter_prefix_values((cid, cindex)).next().is_some() {
```
clangenb, line 1435: I change this line in my debug branch because I think it does the same, but more concise:

```
if <IssuedRewards<T>>::contains_key((cid, cindex), 1) {
```

what do you think?

ceremonies/src/lib.rs
```diff
@@ -1839,11 +1845,15 @@ impl<T: Config> OnCeremonyPhaseChange for Pallet<T> {
...
+					);
 				}
-				let inactives = Self::get_inactive_communities(
-					<encointer_scheduler::Pallet<T>>::current_ceremony_index() - 1,
-					Self::inactivity_timeout(),
+				let inactives = Self::update_inactivity_counters(
+					// check inactivity with delay, such that rewards can be claimed before the inactivity check happens
+					<encointer_scheduler::Pallet<T>>::current_ceremony_index().saturating_sub(2),
+					// compensate for the delay
+					Self::inactivity_timeout().saturating_add(1),
```
clangenb, line 1856: Ah, nice, I see the fix!

However, this kind of seems like we are doing a hack here, maybe we could consider two options?

* we could also just move this to assignment phase
* In the tests I see that we purge at `inactivity_timout + 1`, so from my point of view, we can change the definition here and remove the `saturating_add` to compensate, this would also make it more easy to read.

@brenzi your input?

ceremonies/src/lib.rs
```diff
@@ -1422,7 +1425,7 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
-	fn get_inactive_communities(
+	fn update_inactivity_counters(
```
clangenb, line 1428: I like the renaming! 👍

ceremonies/src/tests.rs
```diff
@@ -1731,31 +1738,128 @@ fn get_inactive_communities_works() {
...
 		assert_eq!(
-			EncointerCeremonies::get_inactive_communities(cindex, timeout, vec![cid0, cid1]),
+			EncointerCeremonies::update_inactivity_counters(cindex, timeout, vec![cid0, cid1]),
 			vec![cid0, cid1]
 		);
 	});
 }
 
+#[test]
+fn purge_inactive_communities_works() {
```
clangenb, line 1767: Nice, heavily needed

clangenb (commented):
One suggestion, looks great otherwise!

ceremonies/src/lib.rs
pifragile: I see what you mean. I was thinking about moving it to the assignment phase but I dont think it makes sense from a UX perspective. Then you could register and suddenly the community is gone.
We could just remove the add. I just wanted to be safe that there are no issues with newly created communities in a setting where the inactivity timeout is 1.
What exactly do you mean with `we can change the definition here and remove the saturating_add to compensate`?

ceremonies/src/lib.rs
pifragile: no it does not the same. we want to check if there is ANY `meetup_index` in the map, not just index 1.

clangenb: But, aren't the meetup indexes starting at 1? Meaning the index 1, will always be populated if there is one meetup?

ceremonies/src/lib.rs
clangenb: However, the inactivity timeout is global, set by encointer governance, hence I suppose we could just disallow 1 in the setter.

Still this is another discussion, we can leave it like that for now, IMO.

ceremonies/src/lib.rs
pifragile: well but it could be that the meetup at index 1 fails and all the others pass, right?

clangenb: Agreed, I am convinced 👍

clangenb (approved):
LGTM, the discussions convinced me. :)

ceremonies/src/lib.rs
brenzi: Isn't the UX even worse now? If we check cindex-2 this means that there may have been 5 idle ceremonies plus one active full ceremony cycle with legit attendees - but still we purge, because of idle 5 ceremonies before the last one.

I think I'd prefer moving this check to the start of assigning phase. No harm if people registered in vain. Big harm if they attended in vain

pifragile: yeah this is true.

brenzi (changes requested):
See comment. Should better check cindex -1 at assigning

brenzi (approved):
Lgtm. We should test manually too before release

PR #248, opened 2022-07-27
Query asset fee details

### LINKED ISSUE #228: add rpc `query_asset_fee_details`
Substrate does not provide RPCs to get fee estimates for custom assets unfortunately. However, this should be very straightforward. We can simply wrap the calls for the `pallet-tx-payment` and apply the balance conversion on the returned values

pifragile:
* Closes #228


balances-tx-payment/rpc/runtime-api/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+# substrate deps
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+frame-support = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-api = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-std = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+
+[features]
+default = ["std"]
+std = [
+    "frame-support/std",
```
clangenb, line 20: `codec/std` is missing.

balances-tx-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+use codec::Codec;
+use encointer_balances_tx_payment_rpc_runtime_api::{
+	BalancesTxPaymentApi as BalancesTxPaymentApiRuntimeApi, Error,
+};
+//use encointer_rpc::Error;
```
clangenb, line 21: obsolete comment

balances-tx-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+		at: Option<BlockHash>,
+	) -> RpcResult<FeeDetails<NumberOrHex>>;
+}
+
+pub struct BalancesTxPaymentRpc<C, P, Q, R, S> {
+	client: Arc<C>,
+	transaction_payment: P,
+	_marker: std::marker::PhantomData<Q>,
+	_marker2: std::marker::PhantomData<R>,
+	_marker3: std::marker::PhantomData<S>,
```
clangenb, line 59: You can rewrite this as `_marker: std::marker::PhantomData<(Q,R,S)>`

balances-tx-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+		+ Encode
+		+ Decode
+		+ Send
+		+ Sync
+		+ PartialEq
+		+ Into<NumberOrHex>
+		+ MaybeDisplay
+		+ Copy,
+	Block: BlockT,
+	P: TransactionPaymentApiServer<<Block as BlockT>::Hash, RuntimeDispatchInfo<Balance>>,
```
clangenb, line 90: Hmmm, why do we need the server here, can't we use the `RuntimeApi` the server uses?

balances-tx-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+
+		let try_into_rpc_balance = |value: AssetBalance| {
+			value.try_into().map_err(|_| {
+				JsonRpseeError::Call(CallError::Custom(ErrorObject::owned(
+					ErrorCode::InvalidParams.code(),
+					format!("{} doesn't fit in NumberOrHex representation", value),
+					None::<()>,
+				)))
+			})
+		};
```
clangenb, line 116: Closures are nice, but as this closure does not need external context, it would be more readable to extract this into a function.

clangenb (commented):
This looks almost as I expected, great! Some minor input though. :)

balances-tx-payment/rpc/src/lib.rs
pifragile: i copied it from here :)
https://github.com/paritytech/substrate/blob/938c0d9363131d862bcc04d2cf913ce9cb9dd5fe/frame/transaction-payment/rpc/src/lib.rs#L146

clangenb: haha, ok.

balances-tx-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+							Some(e.to_string()),
+						))
+					})?
+					.map_err(|_e| {
+						CallError::Custom(ErrorObject::owned(
+							Error::RuntimeError.into(),
+							"Unable to query balance conversion.",
+							Some("Unable to query balance conversion."),
+						))
+					})?;
```
clangenb, line 139: I thought I had written this in the last review, but I seem to have missed it. I would love to see something like this here.

```rust
api.
    .balance_to_asset_balance(&at, inclusion_fee.len_fee, asset_id)
	.map_err(|e| runtime_error_into_rpc_error(e, "Unable to base fee")?;
```

We could re-use the function three times already. :)

balances-tx-payment/rpc/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+				let adjusted_weight_fee = api
+					.balance_to_asset_balance(&at, inclusion_fee.adjusted_weight_fee, asset_id)
+					.map_err(|e| {
+						CallError::Custom(ErrorObject::owned(
+							Error::RuntimeError.into(),
+							"Unable to query fee details.",
+							Some(e.to_string()),
+						))
+					})?
+					.map_err(|_e| {
```
clangenb, line 167: Why are you dropping this error?

clangenb (commented):
Cool that you could do the change to the runtime-api trait bound!

One minor issue and then, we are done. :)

balances-tx-payment/rpc/src/lib.rs
pifragile: ...which is weird because `thiserror` crate claims that

A Display impl is generated for your error if you provide #[error("...")] messages on the struct or each variant of your enum, as shown above in the example.

which we actually do

clangenb (approved):
Nice Looks good to me now!

PR #268, opened 2022-09-12
Reputables can endorse

### LINKED ISSUE #266: allow reputables to endorse N newbies per ceremony
Field experience has shown that it is not realistic to react to sudden spikes of interest in joining the community if only bootstrappers are allowed to endorse newbies.

Observations with Leu community:
 
* some bootstrappers never made use of their right to endorse
* some bootstrappers quickly ran out of endorsement tickets
* some reputables have show to be more effective multipliers than certain bootstrappers
* it didn't seem necessary to expire endorsements after one ceremony. It is too heavy on coordination if the endorses has to first ask the endorsee which ceremony they'd be able to attend. We should just endorse and only clear the endorsement once the endorsee got assigned (so, if noshow, then endorsement is lost)

One way to solve this is leveraging the social graph and empower all reputables to invite a limited number of people per ceremony. We may want to start with a small N (i.e. 1-5)

On the UI side it would make sense to combine the act of sending a minimal amount of Leu with an endorsement into an *invitation* - without confusing people with the details, an invitation would just do both

pifragile:
Closes #266


ceremonies/src/benchmarking.rs
```diff
@@ -340,12 +340,15 @@ benchmarks! {
 
 	endorse_newcomer {
 		let cid = create_community::<T>();
-		let alice: T::AccountId = account("alice", 1, 1);
+		let cindex = encointer_scheduler::Pallet::<T>::current_ceremony_index();
+
+		let zoran= account_id::<T>(&generate_pair());
+		Pallet::<T>::fake_reputation((cid, cindex - 1), &zoran, Reputation::VerifiedUnlinked);
```
clangenb, line 346: Should this benchmark maybe be renamed to `endorse_newcomer_with_reputable`?

ceremonies/src/lib.rs
```diff
@@ -1946,6 +2006,34 @@ impl<T: Config> Pallet<T> {
...
+		}
+		false
+	}
+
+	fn is_endorsed(
+		participant: &T::AccountId,
+		cc: &CommunityCeremony,
+	) -> Option<CeremonyIndexType> {
+		let reputation_lifetime = Self::reputation_lifetime();
+		for i in (0..=reputation_lifetime).rev() {
```
clangenb, line 2028: Big props for the `rev()` for optimization! A comment for the people outside the occult circle to understand would be nice though.

ceremonies/src/tests.rs
```diff
@@ -3042,3 +3175,55 @@ fn attest_attendees_works() {
...
+		Endorsees::<TestRuntime>::insert((cid, 2), &alice, ());
+
+		assert_eq!(EncointerCeremonies::is_endorsed(&alice, &(cid, 4)), Some(2));
+
+		// above reputation lifetime
+		assert_eq!(EncointerCeremonies::is_endorsed(&alice, &(cid, 9)), None);
+
+		Endorsees::<TestRuntime>::insert((cid, 1), &alice, ());
+
+		// will return the earlies endorsement
```
clangenb, line 3226: typo

clangenb (commented):
Nice, it looks very clean! I only have minor comments.

ceremonies/src/lib.rs
pifragile: I think you are giving the props for the wrong thing😅 I use rev in order that the oldest endorsement will be detected, and i didnt have any optimizations in mind. what optimization is it you are talking about?

ceremonies/src/lib.rs
```diff
@@ -486,29 +486,48 @@ pub mod pallet {
...
 
-			<BurnedBootstrapperNewbieTickets<T>>::mutate(&cid, sender.clone(), |b| *b += 1); // safe; limited by AMOUNT_NEWBIE_TICKETS
+			if is_bootstrapper {
+				<BurnedBootstrapperNewbieTickets<T>>::mutate(&cid, sender.clone(), |b| *b += 1);
+			// safe; limited by AMOUNT_NEWBIE_TICKETS
+			} else {
+				<BurnedReputableNewbieTickets<T>>::mutate(&(cid, cindex), sender.clone(), |b| {
+					*b += 1
+				}); // safe; limited by AMOUNT_NEWBIE_TICKETS
+			}
```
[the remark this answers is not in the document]
pifragile: true👍🏽

bazaar/src/benchmarking.rs
```diff
@@ -28,7 +28,7 @@ fn create_community<T: Config>() -> CommunityIdentifier {
 		..Default::default()
 	};
 	encointer_communities::Pallet::<T>::new_community(
-		RawOrigin::Root.into(),
+		RawOrigin::Signed(account("alice", 1, 1)).into(),
```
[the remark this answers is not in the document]
pifragile: ok will open another PR for this

ceremonies/src/lib.rs
clangenb: Haha, well in that case I think that you actually counter optimized. :P Because the closes you are to the current ceremony index, the more likely it is that you have the endorsement for it. I misread the code, so I thought that it checks the newest index first.

Hmm, doesn't the code disallow multiple endorsements?

clangenb: Hmm, as long as we have a small reputation lifetime this is fine, but the dev-network might loop too long if we do it like this. Reversing the order does not resolve that issue however if the endorsee is not registered at all.

In general, I believe we should consider adding a safety parameter that prevents big loops if the reputation lifetime is large. Or, on the other hand, this is why we might want to explicitly pass the reputations after all. But in this specific case, we do not know which ceremony indexes we need to look at. Hmmm, parametrized for loops in the runtime are scary.

For now I suggest we should reverse the order to optimize. Re-thinking for loops we can do at another time.

clangenb (approved):
Looks good to me!

PR #273, opened 2022-10-17
Democracy MVP

pifragile:
This is a working and mergeable  version of the democracy pallet.
There is a [tracking issue](https://github.com/encointer/pallets/issues/347) for potential future work on the pallet.

For details about the protocol please see [this section in the Encointer book](https://book.encointer.org/protocol-democracy.html).


ceremonies/src/lib.rs
```diff
@@ -509,10 +509,7 @@ pub mod pallet {
 			inactivity_timeout: InactivityTimeoutType,
 		) -> DispatchResultWithPostInfo {
 			<T as pallet::Config>::CeremonyMaster::ensure_origin(origin)?;
-			<InactivityTimeout<T>>::put(inactivity_timeout);
-			info!(target: LOG, "set inactivity timeout to {}", inactivity_timeout);
-			Self::deposit_event(Event::InactivityTimeoutUpdated(inactivity_timeout));
-			Ok(().into())
+			Self::do_set_inactivity_timeout(inactivity_timeout)
```
brenzi, line 512: I'm not a fan of scattering few lines of code into functions that are called just in one place. This was more readable before IMO

ceremonies/src/lib.rs
```diff
@@ -1754,6 +1772,8 @@ impl<T: Config> Pallet<T> {
 					participant,
 					Reputation::VerifiedUnlinked,
 				);
+				<ReputationCount<T>>::mutate((&cid, cindex), |b| *b += 1); // safe, as reputation_count is limited by the number of locations available on earth
```
brenzi, line 1775: is there a reason why you mutate every single issuance instead of counding issuances in your loop and mutating once? db access is expensive and should be minimized

ceremonies/src/lib.rs
```diff
@@ -1900,10 +1920,24 @@ impl<T: Config> Pallet<T> {
...
+		if cindex < current_cindex.saturating_sub(Self::reputation_lifetime()) {
+			return false
+		}
+		<ParticipantReputation<T>>::get((*cid, cindex), account_id).is_verified()
+	}
+
+	#[cfg(any(test, feature = "runtime-benchmarks", feature = "mocks"))]
 	// only to be used by tests
-	fn fake_reputation(cidcindex: CommunityCeremony, account: &T::AccountId, rep: Reputation) {
+	pub fn fake_reputation(cidcindex: CommunityCeremony, account: &T::AccountId, rep: Reputation) {
```
brenzi, line 1937: this looks helpful. we may even consider exposing this as a root origin extrinsic on testnets

communities/src/lib.rs
```diff
@@ -298,18 +298,7 @@ pub mod pallet {
...
-				.map_err(|_| <Error<T>>::InvalidNominalIncome)?;
-			Self::ensure_cid_exists(&cid)?;
-
-			<NominalIncome<T>>::insert(cid, nominal_income);
-
-			info!(target: LOG, " updated nominal income for cid: {:?}", cid);
-			Self::deposit_event(Event::NominalIncomeUpdated(cid, nominal_income));
-
-			Ok(().into())
+			Self::do_update_nominal_income(cid, nominal_income)
```
brenzi, line 301: same argument as above. few lines of code should stay in the dispatchable

democracy/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+name = "pallet-encointer-democracy"
+version = "1.2.0"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.6.1", default-features = false, features = ["derive"] }
+log = { version = "0.4.14", default-features = false }
+scale-info = { version = "2.5.0", default-features = false }
+
+# local deps
+encointer-balances = { package = "pallet-encointer-balances", path = "../balances", default-features = false }
```
brenzi, line 13: I suspect this could move to dev-dependencies?

democracy/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+				.saturating_add(T::ProposalLifetimeCycles::get()))..=
+				(proposal.start_cindex.saturating_sub(2)))
+				.collect::<Vec<CeremonyIndexType>>())
+		}
+		/// Returns the reputations that
+		/// 1. are valid
+		/// 2. have not been used to vote for proposal_id
+		/// 3. originate in the correct community (for Community AccessPolicy)
+		/// 4. are within proposal.start_cindex - reputation_lifetime + proposal_lifetime and proposal.start_cindex - 2
+		pub fn valid_reputations(
```
brenzi, line 219: reading your fn doc above, I could imagine that ` eligible_reputations`
would be a more precise naming.

brenzi: question: If ppl change their minds, this logic will prevent changing the vote of already committed reputation, right?

democracy/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		ProposalEnacted(ProposalIdType),
+	}
+
+	#[pallet::error]
+	#[derive(PartialEq, Eq)]
+	pub enum Error<T> {
```
brenzi, line 76: please add doc strings with `///`

democracy/src/tests.rs
```diff
@@ -0,0 +1,684 @@
...
+		assert_eq!(EncointerDemocracy::is_passing(1).unwrap(), true);
+
+		// high turnout 90 % approval
+		Tallies::<TestRuntime>::insert(1, Tally { turnout: 100, ayes: 90 });
+		assert_eq!(EncointerDemocracy::is_passing(1).unwrap(), true);
+	});
+}
+
+#[test]
+fn enactment_updates_proposal_metadata_andenactment_queue() {
```
brenzi, line 601: missing underscore

democracy/src/tests.rs
```diff
@@ -0,0 +1,684 @@
...
+		assert_eq!(EncointerDemocracy::proposals(2).unwrap().state, ProposalState::Enacted);
+
+		assert_eq!(EncointerDemocracy::enactment_queue(proposal_action.get_identifier()), None);
+
+		assert_eq!(EncointerDemocracy::enactment_queue(proposal_action2.get_identifier()), None);
+	});
+}
+
+#[test]
+fn enact_update_nominal_income_works() {
```
brenzi, line 637: this is the main happy flow test from proposal to enactment. I'd prefer if there was no state faking in this test but real non-unanimous voting action with positive outcome

brenzi (changes requested):
this looks very clean to me. only a few easy-to-fix remarks

democracy/src/tests.rs
pifragile: generally i think that it is important to have small **unit** tests that cover one specific aspect of the functionality, the rest can be mocked away for efficiency. in this case we will have enactment unit tests for every possible `ProposalAction` and it would no be efficient IMO to build the whole setup for each of those tests.

nevertheless the input to have a complete happy flow test was really good, as after adding one i discovered that i forgot to update the proposal state on voting which basically makes the whole thing not do anything😂 fixed this now and also added a lazy proposal update extrinsic for the case where everyone has voted and we still need to update the proposal state.

ceremonies/src/lib.rs
pifragile: yes, makes total sense👍🏽

democracy/src/lib.rs
pifragile: renamed. yes no changing minds in the current protocol

ceremonies/src/lib.rs
brenzi: Ah, didn't see that. Perfect argument

democracy/src/lib.rs
brenzi: that's fine. can be declared as such. and I'm not even sure we want to change that

brenzi (approved):

PR #274, opened 2022-10-25
Bump substrate v0.9.28 and add a `MeetupResult` in the `IssuedRewards` storage.

echevrier:
Upgrade to polkadot-v0.9.28 :
- substrate dependencies to substrate's branch = "polkadot-v0.9.28"
- polkadot dependencies to polkadot's branch = "release-v0.9.28"
- jsonrpsee to version = 0.15.1

### Reasoning for adding the `MeetupResult`:
Failing calls do no longer change the storage, see https://github.com/encointer/pallets/issues/275. So we had to change the dispatch result to be OK, but in turn we store the `MeetupResult` in `IssuedRewards` instead of just a plain `()`. This is anyhow helpful information for client applications. Further, an event is emitted when a failing meetup is evaluated.


primitives/src/ceremonies.rs
```diff
@@ -347,6 +348,22 @@ pub struct ReputationCacheValue {
...
 }
 
+#[derive(Encode, Decode, Copy, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "serde_derive", derive(Serialize, Deserialize))]
+pub enum MeetupResult {
+	Ok,
+	VotesNotDependable,
+	MeetupValidationIndexOutOfBounds,
+	NoResult,
+	Other,
```
clangenb, line 358: If these two are not used, I would remove them. `NoResult` I would never use. If this is semantically correct somewhere, I would rather go with an `Option<MeetupResult>` and then set it to `None`.

primitives/src/ceremonies.rs
```diff
@@ -347,6 +348,22 @@ pub struct ReputationCacheValue {
...
+}
+
+impl Default for MeetupResult {
+	fn default() -> Self {
+		MeetupResult::NoResult
+	}
+}
```
clangenb, line 365: If we remove the two unused variants above, I think we remove the default implementation.

ceremonies/src/lib.rs
```diff
@@ -611,15 +608,31 @@ pub mod pallet {
...
+							)
 						},
+					};
+					// only mark issuance as complete in registering phase
+					// because in attesting phase there could be a failing early payout attempt
+					if current_phase == CeremonyPhaseType::Registering {
+						info!(target: LOG, "marking issuance as completed for failed meetup.");
+
+						<IssuedRewards<T>>::insert((cid, cindex), meetup_index, meetup_result);
+						return Ok(Pays::No.into())
```
clangenb, line 633: We should also return the error here. Otherwise, the user expects some rewards.

clangenb (changes requested):
Some minor comments. Otherwise, it looks good.

@pifragile you might also want to have a look at this.

Further, I am considering of changing the paradigm, such that we also have `branch = polkadot-v9.x.x` on master. Do you have any objections here?

ceremonies/src/lib.rs
echevrier: If we return an error, the storage (IssuedRewards) will not be updated. We did these changes to fix it: https://github.com/encointer/pallets/issues/275
The "error" feedback is in the MeetupResult:
MeetupResult::VotesNotDependable or MeetupResult::MeetupValidationIndexOutOfBounds. Perhaps we can improve the name:  MeetupResult::VotesNotDependableError, MeetupResult::MeetupValidationIndexOutOfBoundsError ?

clangenb: Ah, sorry, I completely forgot that this was the cause of this PR. In that case, we should additionally add an event:

`MeetupEvaluated(MeetupResult)`

ceremonies/src/lib.rs
```diff
@@ -1156,7 +1168,7 @@ pub mod pallet {
 		CommunityCeremony,
 		Blake2_128Concat,
 		MeetupIndexType,
-		MeetupResult,
+		Option<MeetupResult>,
 		ValueQuery,
```
clangenb, line 1172: Instead of making this an `Option<MeetupResult>`, you should change `ValueQuery` to `OptionQuery`.

clangenb (commented):
One tiny remark, which will lead you to revert some stuff here. :)

ceremonies/src/tests.rs
```diff
@@ -1035,9 +1037,15 @@ fn claim_rewards_error_results_in_meetup_marked_as_completed() {
...
-			.is_err());
+			.is_ok());
 		// in registering, the meetup is marked as completed
 		assert!(IssuedRewards::<TestRuntime>::contains_key((cid, cindex), 1));
+		let meetup_result = IssuedRewards::<TestRuntime>::get((cid, cindex), 1);
+		assert_eq!(meetup_result, Some(MeetupResult::VotesNotDependable));
+
+		assert!(event_deposited::<TestRuntime>(
+			Event::MeetupEvaluated(cid, 1, MeetupResult::VotesNotDependable).into()
+		));
```
clangenb, line 1048: 👍

clangenb (approved):
Thanks, looks good to me now!

pifragile:
sorry i completely missed this @clangenb

PR #281, opened 2022-11-15
fix get_meetup_participants

pifragile:


ceremonies/src/tests.rs
```diff
@@ -950,6 +950,118 @@ fn claim_rewards_works_with_one_missing_attestation() {
...
+
+		assert_err!(
+			EncointerCeremonies::claim_rewards(
+				Origin::signed(account_id(&all_participants[0].clone())),
+				cid,
+				Some(2 * meetup_count + 1)
+			),
+			Error::<TestRuntime>::InvalidMeetupIndex,
+		);
+	});
```
clangenb, line 1062: Nice, thanks for the extensive tests. I am convinced that the error is fixed. 👍

clangenb (approved):

PR #283, opened 2022-11-19
Remove `attest_claims` extrinsic

### LINKED ISSUE #280: Remove attest_claims extrinsic
Since the app uses attest_attendees now, we should get rid of this legacy extrinsic.

We realized that the two are in fact incompatible if both are used by attendees of the same meetup, because one registers the vote of origin an the other registers votes of attestees, leaving an edge case where a single attendee with the old version will not have their vote registered and not get a reward

https://github.com/encointer/pallets/blob/19fbb18fff2a326b88dc3127c6c154a4ea0366ef/ceremonies/src/lib.rs#L310

Must bump tx version in runtime!

clangenb:
* Closes #280 

Changes:
* removed `attest_claims` extrinsic.
* Removed all tests that simply checked that the claim contains valid data.
* All other tests that were testing that the meetup constellation works have simply been adjusted to use the new `attest_attendees` extrinsic.
* removed obsolete `attest_claims` weight.
* Some clean-up.


ceremonies/src/tests.rs
```diff
@@ -788,20 +542,18 @@ fn claim_rewards_works() {
...
-			Event::NoReward {
+			BalancesEvent::Issued(
 				cid,
-				cindex,
-				meetup_index: 1,
-				account: account_id(&ferdie),
-				reason: ExclusionReason::TooFewOutgoingAttestations,
-			}
+				charlie.clone(),
+				EncointerCeremonies::ceremony_reward().lossy_into()
```
clangenb, line 555: Charlie does now receive rewards because he supplies valid attestations. Before, he handed in claims with invalid signatures.

ceremonies/src/tests.rs
```diff
@@ -810,8 +562,8 @@ fn claim_rewards_works() {
 				cid,
 				cindex,
 				meetup_index: 1,
-				account: account_id(&eve),
-				reason: ExclusionReason::TooFewOutgoingAttestations,
+				account: ferdie.clone(),
+				reason: ExclusionReason::NoVote,
```
clangenb, line 566: Before they had votes, as other people supplied claims for them, even though ferdie and eve did not send any attestations. Now the other attendees don't hand in votes for them anymore, which is why the error changed.

brenzi (approved):
massive cleanup. thank you
haven't tested anything, but as this forces us to adjust integration tests anyway, I won't hold up this PR

PR #288, opened 2022-11-26
[ceremonies] bootstrappers can also endorse as reputables

### LINKED ISSUE #287: bootstrappers should get reputable endorsement tickets once BS tickets exhausted
Currently, bootstrappers get a one-off allowance for endorsements which does not renew. Reputables get new endorsement tickets for every cycle

This does not make sense if we assume BS to be very active endorsers even after the initial growth phase of the first few cycles

The logic should be:
```
if bootstrapper
  if BS tickets left
    endorse
  else if reputabletickets left
    endorse
  else
    error
```

clangenb:
* Closes #287 

As commented in #287 I think it makes more sense to first burn the reputable newbie tickets and then the bootstrapper newbie tickets.

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -332,26 +332,7 @@ pub mod pallet {
 				Error::<T>::AlreadyEndorsed
 			);
 
-			if <encointer_communities::Pallet<T>>::bootstrappers(&cid).contains(&sender) {
-				ensure!(
-					<BurnedBootstrapperNewbieTickets<T>>::get(&cid, &sender) <
-						Self::endorsement_tickets_per_bootstrapper(),
-					Error::<T>::NoMoreNewbieTickets
-				);
-				<BurnedBootstrapperNewbieTickets<T>>::mutate(&cid, sender.clone(), |b| *b += 1);
-			// safe; limited by AMOUNT_NEWBIE_TICKETS
-			} else if Self::has_reputation(&sender, &cid) {
-				ensure!(
-					<BurnedReputableNewbieTickets<T>>::get(&(cid, cindex), &sender) <
-						Self::endorsement_tickets_per_reputable(),
-					Error::<T>::NoMoreNewbieTickets
-				);
-				<BurnedReputableNewbieTickets<T>>::mutate(&(cid, cindex), sender.clone(), |b| {
-					*b += 1
-				}); // safe; limited by AMOUNT_NEWBIE_TICKETS
-			} else {
-				return Err(Error::<T>::AuthorizationRequired.into())
-			}
+			Self::burn_newbie_tickets(cid, cindex, &sender)?;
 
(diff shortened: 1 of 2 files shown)


ceremonies/src/tests.rs
```diff
@@ -1452,15 +1452,15 @@ fn endorse_newbie_fails_if_already_endorsed_in_previous_ceremony() {
...
-fn endorse_newbie_fails_if_not_authorized() {
+fn endorse_newbie_fails_without_reputation() {
 	new_test_ext().execute_with(|| {
 		let cid = perform_bootstrapping_ceremony(None, 1);
 
 		let yran = account_id(&sr25519::Pair::from_entropy(&[8u8; 32], None).0);
 		let zoran = sr25519::Pair::from_entropy(&[9u8; 32], None).0;
 		assert_err!(
 			EncointerCeremonies::endorse_newcomer(Origin::signed(account_id(&zoran)), cid, yran),
-			Error::<TestRuntime>::AuthorizationRequired
```
brenzi, line 1463: AuthorizationRequired is now an orphaned Error. Shall we remove it (I think there are other orphans by now)
This will mess up the indexes of the enum, but I think we always rely on metadatsa, so that shouldn't matter ?

ceremonies/src/tests.rs
```diff
@@ -1452,15 +1452,15 @@ fn endorse_newbie_fails_if_already_endorsed_in_previous_ceremony() {
 }
 
 #[test]
-fn endorse_newbie_fails_if_not_authorized() {
+fn endorse_newbie_fails_without_reputation() {
```
brenzi, line 1455: well, bootstrappers initially have no reputation, so the naming might mislead.

ceremonies/src/lib.rs
```diff
@@ -1278,6 +1259,36 @@ impl<T: Config> Pallet<T> {
 			<NewbieIndex<T>>::contains_key((cid, cindex), &sender)
 	}
 
+	/// Will burn the `sender`'s newbie tickets if he has some.
+	///
+	/// First we try to use the the reputable tickets because they refill with new reputation, and
+	/// then we try to use the bootstrapper tickets.
+	fn burn_newbie_tickets(
```
brenzi, line 1266: nice solution

brenzi (approved):
no blockers here, just remarks

ceremonies/src/tests.rs
clangenb: Yes, it should not matter. I believe it does not even matter for the rust-cli any more. I think we can remove it.

ceremonies/src/tests.rs
clangenb: Yes, I renamed it: `endorse_newbie_fails_if_sender_has_no_reputation_and_is_not_bootstrapper`

PR #289, opened 2022-11-26
types and doc cleanup

brenzi:
a few little things that bugged me when reviewing the new endorsement code

I don't think it is good practise to use base types in pallet storage if used in more than one place


clangenb (commented):
Yes, the u8 has bugged me so hard!

clangenb (approved):

PR #290, opened 2022-11-26
extend unit test coverage

brenzi:


ceremonies/src/tests.rs
```diff
@@ -3153,10 +3153,25 @@ fn has_reputation_works() {
...
 		assert_eq!(EncointerCeremonies::has_reputation(&alice, &cid), false);
 
+		// reputation type does not qualify
+		EncointerCeremonies::fake_reputation((cid, 1), &alice, Reputation::Unverified);
+
+		assert_eq!(EncointerCeremonies::has_reputation(&alice, &cid), false);
+
+		EncointerCeremonies::fake_reputation((cid, 1), &alice, Reputation::UnverifiedReputable);
+
+		assert_eq!(EncointerCeremonies::has_reputation(&alice, &cid), false);
```
brenzi, line 3168: I actually wonder if we shouldn't accept `UnverifiedReputable` as `has_reputation`, because the account is just an alias and the controller of the account has reputation (even if rep is on another account.

@pifragile ?

ceremonies/src/tests.rs
```diff
@@ -1092,7 +996,7 @@ fn early_rewards_with_new_attest_attendees_extrinsic_with_one_missing_attestatio
 }
 
 #[test]
-fn early_rewards_does_not_work_with_one_missing_attestation() {
```
brenzi, line 1095: the previous testing logic didn't cover the test scenario I expected from the name. I think it makes more sense to test a missing submission of attestations. alternatively, we could keep both tests, but I don't think this is necessary

ceremonies/src/tests.rs
```diff
@@ -1009,48 +954,7 @@ fn early_rewards_works() {
 }
 
 #[test]
-fn early_rewards_with_new_attest_attendees_extrinsic_works() {
```
brenzi, line 1012: this entire test is redundant since we removed the old extrinsic and adjusted the other test to cover this

ceremonies/src/lib.rs
```diff
@@ -684,7 +683,7 @@ pub mod pallet {
 		/// meetup time calculation failed
 		MeetupTimeCalculationError,
 		/// no valid claims were supplied
-		NoValidClaims,
+		NoValidAttestations,
```
brenzi, line 686: with changed attestation logic, this name has to change as well

ceremonies/src/lib.rs
```diff
@@ -278,7 +278,6 @@ pub mod pallet {
 			let (cindex, meetup_index, meetup_participants, _meetup_location, _meetup_time) =
 				Self::gather_meetup_data(&cid, &sender)?;
 
-			ensure!(meetup_participants.contains(&sender), Error::<T>::OriginNotParticipant);
```
brenzi, line 281: this ensure is futile because the meetup is derived from the participant and the above line would fail if there a^was no assignment for the sender

ceremonies/src/tests.rs
```diff
@@ -415,7 +421,7 @@ fn attest_attendees_for_non_participant_fails_silently() {
 }
 
 #[test]
-fn attest_attendee_for_non_participant_fails() {
```
brenzi, line 418: this test didn't test what it claimed to test.

I now made two out of one, to trigger both possible errors

ceremonies/src/tests.rs
pifragile: I understand your reasoning but I'm not sure, as like this reputations would be counted twice. You can have reputation with account A, then use this reputation to register account B for a gathering. B then becomes `UnverifiedReputable`. Then both accounts A and B could endorse newbies.

ceremonies/src/tests.rs
```diff
@@ -753,49 +759,18 @@ fn claim_rewards_can_only_be_called_for_valid_meetup_indices() {
...
-
-		assert_err!(
-			EncointerCeremonies::claim_rewards(
-				Origin::signed(account_id(&all_participants[0].clone())),
-				cid,
-				Some(2 * meetup_count + 1)
-			),
-			Error::<TestRuntime>::InvalidMeetupIndex,
-		);
+		for index in
```
pifragile, line 762: nice :)

ceremonies/src/tests.rs
pifragile: i think we should keep both tests, because those are two completely different things.
the old test could be renamed to `early_rewards_does_not_work_with_one_missing_attestation_per_submitter`

ceremonies/src/tests.rs
brenzi: True. No harm done, but still it may make sense to leave it as is

ceremonies/src/tests.rs
brenzi: Agreed

pifragile (approved):
LGTM

PR #298, opened 2022-12-27
Make demurrage computation sound

clangenb:
Fixes potential overflows when computing the demurrage factor, see: 
* https://github.com/encointer/encointer-node/issues/290
* Closes https://github.com/encointer/pallets/pull/297

### DIFF
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -19,7 +19,7 @@
 pub use crate::weights::WeightInfo;
 use core::marker::PhantomData;
 use encointer_primitives::{
-	balances::{BalanceEntry, BalanceType, Demurrage, DemurrageError, FeeConversionFactorType},
+	balances::{BalanceEntry, BalanceType, Demurrage, FeeConversionFactorType},
 	communities::CommunityIdentifier,
 };
 use frame_support::{
@@ -251,18 +251,7 @@ impl<T: Config> Pallet<T> {
 	) -> BalanceEntry<T::BlockNumber> {
 		let current_block = frame_system::Pallet::<T>::block_number();
 
-		match entry.apply_demurrage(demurrage, current_block) {
-			Ok(updated_entry) => updated_entry,
-			Err(e) => {
-				// This should never happen in production! We handle every potential error case
-				// explicitly to be sure  that we detect changes in the `DemurrageError` enum.
-
-				match e {
-					DemurrageError::DemurrageMustBePositive =>
-						unreachable!("Demurrage is always positive; qed"),
-				}
-			},
-		}
+		entry.apply_demurrage(demurrage, current_block)
 	}
 
 	/// Create a new account on-chain if it does not exist.
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -258,11 +258,6 @@ impl<T: Config> Pallet<T> {
 				// explicitly to be sure  that we detect changes in the `DemurrageError` enum.
 
 				match e {
-					DemurrageError::LastBlockBiggerThanCurrent =>
-						unreachable!("block number monotonically increases; qed"),
-					DemurrageError::ElapsedBlocksMoreThan32Bits => unreachable!(
-						"This can only be reached around the year 2720 with 6 seconds block time; qed"
-					),
 					DemurrageError::DemurrageMustBePositive =>
 						unreachable!("Demurrage is always positive; qed"),
 				}
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -254,10 +254,15 @@ impl<T: Config> Pallet<T> {
 		match entry.apply_demurrage(demurrage, current_block) {
 			Ok(updated_entry) => updated_entry,
 			Err(e) => {
-				// This should never happen in production!
+				// This should never happen in production! The `apply_demurrage` method is designed
+				// to only return an error when some blockchain invariants no longer hold like:
+				// * last_update < current_block
+				// * demurrage < 0
+				// * blocknumber > 2^32
+				//
+				// Still if such an invariant is violated, what should we do here? Shall we use an
+				// expect after all?
 				debug!("Error when applying demurrage: {:?}", e);
-
-				// Todo: is this the safe fallback, or shall we introduce expects?
 				Default::default()
 			},
 		}
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -247,7 +247,7 @@ impl<T: Config> Pallet<T> {
 	///     * exp(-1* demurrage_rate_per_block * number_of_blocks_since_last_written)
 	pub(crate) fn apply_demurrage(
 		entry: BalanceEntry<T::BlockNumber>,
-		demurrage: BalanceType,
+		demurrage: Demurrage,
 	) -> BalanceEntry<T::BlockNumber> {
 		let current_block = frame_system::Pallet::<T>::block_number();
 
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -20,7 +20,7 @@ pub use crate::weights::WeightInfo;
 use core::marker::PhantomData;
 use encointer_primitives::{
 	balances::{BalanceEntry, BalanceType, Demurrage, FeeConversionFactorType},
-	communities::CommunityIdentifier,
+	communities::{validate_demurrage, CommunityIdentifier, RangeError},
 };
 use frame_support::{
 	dispatch::DispatchResult,
@@ -367,8 +367,13 @@ impl<T: Config> Pallet<T> {
 		<DemurragePerBlock<T>>::try_get(cid).unwrap_or_else(|_| T::DefaultDemurrage::get())
 	}
 
-	pub fn set_demurrage(cid: &CommunityIdentifier, demurrage: Demurrage) {
+	pub fn set_demurrage(
+		cid: &CommunityIdentifier,
+		demurrage: Demurrage,
+	) -> Result<(), RangeError> {
+		validate_demurrage(&demurrage)?;
 		<DemurragePerBlock<T>>::insert(cid, &demurrage);
+		Ok(())
 	}
 
 	pub fn purge_balances(cid: CommunityIdentifier) {
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -21,7 +21,6 @@ use core::marker::PhantomData;
 use encointer_primitives::{
 	balances::{BalanceEntry, BalanceType, Demurrage, FeeConversionFactorType},
 	communities::CommunityIdentifier,
-	fixed::transcendental::exp,
 };
 use frame_support::{
 	dispatch::DispatchResult,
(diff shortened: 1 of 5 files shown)


balances/src/lib.rs
```diff
@@ -251,21 +250,21 @@ impl<T: Config> Pallet<T> {
...
+				// to only return an error when some blockchain invariants no longer hold like:
+				// * last_update < current_block
+				// * demurrage < 0
+				// * blocknumber > 2^32
+				//
+				// Still if such an invariant is violated, what should we do here? Shall we use an
+				// expect after all?
+				debug!("Error when applying demurrage: {:?}", e);
+				Default::default()
+			},
```
clangenb, line 267: Here, I honestly don't know what should be the way to handle it. As the function is used inside the fungibles' implementation, we can't just forward an error, so we either panic here or we swallow the error and return something.

primitives/src/balances.rs
```diff
@@ -56,6 +57,90 @@ pub struct BalanceEntry<BlockNumber> {
...
+	// inverse of these operations at the end, which is why we can set the demurrage factor to
+	// 0 in this case.
+
+	let exponent = match (-demurrage_per_block).checked_mul(elapsed_blocks.into()) {
+		Some(exp) => exp,
+		None => return Ok(0.into()),
+	};
+
+	Ok(exp(exponent).unwrap_or(0.into()))
+}
```
clangenb, line 131: This is the handling that fixes the runtime panic. We did an unwrap here at line 130.

balances/src/lib.rs
```diff
@@ -251,21 +250,23 @@ impl<T: Config> Pallet<T> {
...
 	) -> BalanceEntry<T::BlockNumber> {
 		let current_block = frame_system::Pallet::<T>::block_number();
-		let elapsed_time_block_number = current_block - entry.last_update;
-		let elapsed_time_u32: u32 = elapsed_time_block_number
-			.try_into()
-			.ok()
-			.expect("blockchain will not exceed 2^32 blocks; qed");
-		let elapsed_time = BalanceType::from_num(elapsed_time_u32);
-		let exponent: BalanceType = -demurrage * elapsed_time;
-		let exp_result: BalanceType = exp(exponent).unwrap();
```
clangenb, line 261: This panicked because the `exp` function computes the result by flipping the negative sign, and inverting the result at the end again.

primitives/src/balances.rs
```diff
@@ -56,6 +57,91 @@ pub struct BalanceEntry<BlockNumber> {
...
+		if demurrage_per_block < 0 {
+			return Err(DemurrageError::DemurrageMustBePositive)
+		}
+
+		if self.last_update == current_block {
+			// Nothing to be done, as no time elapsed.
+			return Ok(self)
+		}
+
+		if self.principal.eq(&0i16) {
```
haerdib, line 90: Is it worth doing this check every time?
E.g. what's more often the case: `principal ==` 0 or `principal !=0`? If it's the latter, isn't it a waste of time?

primitives/src/balances.rs
```diff
@@ -56,6 +57,91 @@ pub struct BalanceEntry<BlockNumber> {
...
+	/// The following formula is applied to the principal:
+	///    updated_principal = old_principal * e^(-demurrage_per_block * elapsed_blocks)
+	///
+	/// **Note**: This function will be used at every single transaction that is paid with community
+	/// currency. It is important that it is as efficient as possible, but also that it is bullet-
+	/// proof (no-hidden panics!!).
+	pub fn apply_demurrage(
+		self,
+		demurrage_per_block: Demurrage,
+		current_block: BlockNumber,
```
haerdib, line 79:
```suggestion
		current_block_number: BlockNumber,
```

primitives/src/balances.rs
```diff
@@ -56,6 +57,92 @@ pub struct BalanceEntry<BlockNumber> {
...
+}
+
+/// e^(-demurrage_per_block * elapsed_blocks) within [0,1).
+///
+/// Returns an error if the demurrage is negative.
+pub fn demurrage_factor(
+	demurrage_per_block: Demurrage,
+	elapsed_blocks: u32,
+) -> Result<BalanceType, DemurrageError> {
+	if demurrage_per_block < 0 {
```
haerdib, line 120: That's the second check if the demurrage is negative. Remove one?

primitives/src/balances.rs
```diff
@@ -56,6 +57,92 @@ pub struct BalanceEntry<BlockNumber> {
...
+		return Err(DemurrageError::DemurrageMustBePositive)
+	}
+
+	// We can only have errors here if one of the operations overflowed. However, we take the
+	// inverse of these operations at the end, which is why we can set the demurrage factor to
+	// 0 in this case.
+
+	// demurrage > 0; hence exponent <= 0
```
haerdib, line 128:
```suggestion
	// demurrage >= 0; hence exponent <= 0
```

primitives/src/balances.rs
```diff
@@ -56,6 +57,92 @@ pub struct BalanceEntry<BlockNumber> {
...
+pub fn demurrage_factor(
+	demurrage_per_block: Demurrage,
+	elapsed_blocks: u32,
+) -> Result<BalanceType, DemurrageError> {
+	if demurrage_per_block < 0 {
+		return Err(DemurrageError::DemurrageMustBePositive)
+	}
+
+	// We can only have errors here if one of the operations overflowed. However, we take the
+	// inverse of these operations at the end, which is why we can set the demurrage factor to
```
haerdib, line 125: "inverse" only makes sense if one knows about what the `exp` function does. There is no inverse in this function.

How about:
> However, if the calculation overflows, the exponent is too big to calculate. In this case (because e^(-x) -> 0 for big x) we can approximate to 0.

haerdib (commented):
Awesome, that's some documentation ! ❤️

primitives/src/balances.rs
clangenb: `principal == 0` it should not be more often, but at this level we don't know. The overflow bug we had, was on a `BalanceEntry::default()`, which was created when an account did not exist, and then the demurrage was applied since genesis. Of course the overflow error is fixed now, but it is just so much cheaper to make this check here more often than to perform the entire demurrage computation in vain.

primitives/src/balances.rs
clangenb: I just never want this to panic again, so I don't want to make assumptions about the checks on the outside here.

balances/src/lib.rs
```diff
@@ -251,21 +250,23 @@ impl<T: Config> Pallet<T> {
 		demurrage: BalanceType,
```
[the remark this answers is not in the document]
clangenb: You are absolutely right, good catch!

primitives/src/balances.rs
clangenb: Marvellous catch!

primitives/src/balances.rs
```diff
@@ -110,11 +197,60 @@ impl Convert<u128, BalanceType> for EncointerBalanceConverter {
...
+	fn apply_demurrage_when_demurrage_is_negative_errs() {
+		let bal = BalanceEntry::<u32>::new(0.into(), 0);
+		assert_eq!(
+			bal.apply_demurrage(Demurrage::from_num(-0.5), ONE_YEAR).unwrap_err(),
+			DemurrageError::DemurrageMustBePositive,
+		);
+	}
+
+	#[test]
+	fn apply_demurrage_with_overflowing_values_works() {
```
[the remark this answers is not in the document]
clangenb: Absolutely, thanks for the hints! This is why it is good to always have another pair of eyes. :)

balances/src/lib.rs
clangenb: I added very explicit unreachable for all the errors, which make it clear at pallet level that it should never panic.

brenzi (commented):
LGTM. But: 
rescuing any error to zero assumes that an overflow due to very high demurrage is the only source of error. However, in `exp` I can spot many places where an error can be thrown:

https://github.com/encointer/substrate-fixed/blob/a4fb461aae6205ffc55bed51254a40c52be04e5d/src/transcendental.rs#L254-L277

I wonder if we should be more precise with errors in `substrate-fixed` and then handle each case separately with an acceptable fallback. 

I can't think of any other reason than the one you've caught now. But returning `0` is only acceptable if your assumptions are correct

primitives/src/balances.rs
```diff
@@ -56,6 +61,89 @@ pub struct BalanceEntry<BlockNumber> {
...
+	// We can only have errors here if one of the operations overflowed.
+	//
+	// However, as we compute exp(-x), which goes to 0 for big x, we can
+	// approximate the result as 0 if we overflow somewhere on the way
+	// because of the big x.
+
+	// demurrage >= 0; hence exponent <= 0
+	let exponent = match (-demurrage_per_block.abs()).checked_mul(elapsed_blocks.into()) {
+		Some(exp) => exp,
+		None => return 0.into(),
```
brenzi, line 140: zero would cause exp(0) to be 1 which would be returned as demurrage_factor. However, if checked_mul would overflow (which can't happen with sound demurrage values) we should rather skip exp and return 0 as a result, because elapsed_blocks is very high, so 0 is more reasonable than 1.

brenzi, line 140: zero would cause exp(0) to be 1 which would be returned as demurrage_factor. However, if checked_mul would overflow (which can't happen with sound demurrage values) we should rather skip exp and return 0 as a result, because elapsed_blocks is very high, so 0 is more reasonable than 1.

brenzi, line 140: zero would cause exp(0) to be 1 which would be returned as demurrage_factor. However, if checked_mul would overflow (which can't happen with sound demurrage values) we should rather skip exp and return 0 as a result, because elapsed_blocks is very high, so 0 is more reasonable than 1.

primitives/src/balances.rs
```diff
@@ -56,6 +61,89 @@ pub struct BalanceEntry<BlockNumber> {
...
+	};
+
+	// exponent <= 0; hence return value [0, 1)
+	exp(exponent).unwrap_or_else(|_| 0.into())
```
brenzi, line 144: as commented above, this assumes that "too much demurrage" is the only possible source of error. is it?

primitives/src/balances.rs
clangenb: Good catch! But the return 0 here does really return 0 from the function. So it already is, as you say. :)

primitives/src/balances.rs
clangenb: Yes, but this is the case, I have dug into the `exp` function. I have commented this a few lines before, but maybe it is not clear that this also counts for this one. I can adapt the documentation.

clangenb: Yes, but this is the case for all relevant computations. I have dug into the `exp` function. It is not very obvious, but it can only error if:

* The number of fractional bits is 0.
* Or it overflows somewhere, where the overflows can either hapen if either:
   * exponent * intermediate results overflows
   * additions of intermediate results overflow
   * division by 0, when computing `exp(-x)=1/exp(x)`

clangenb:
Yeah I agree about the nothing saying errors in substrate-fixed, which are not ideal. On some occasions, we can even prove with substrate-fixed internal logic that an error can't occur there, which would again simplify the handling on the outside.

primitives/src/balances.rs
clangenb: If this weren't the case, this test would not pass: https://github.com/encointer/pallets/blob/681154a785b70c039fb6a8a9fd24fc1166be2505/primitives/src/balances.rs#L279

brenzi: Ah, misread the code. All good

brenzi (approved):

primitives/src/balances.rs
```diff
@@ -56,6 +61,89 @@ pub struct BalanceEntry<BlockNumber> {
...
+			.checked_mul(demurrage_factor)
+			.expect("demurrage_factor [0,1), hence can't overflow; qed");
+
+		Self { principal, last_update: current_block_number }
+	}
+}
+
+/// e^(-demurrage_per_block * elapsed_blocks) within [0,1).
+///
+/// It will take the absolute value of the `demurrage_per_block` if it is negative.
```
pifragile, line 129: It is a detail but to me it seems more logical to use `max(demurrage_per_block, 0)` than `demurrage_per_block.abs()`

clangenb: For me the idea was that maybe `e^-demurrage*block` is unclear and that they accidentally include the `-` into the demurrage.

pifragile: ok, also a valid point :)

pifragile:
lgtm now

primitives/src/balances.rs
clangenb: Negative demurrage can be handled now anyhow, so we don't have to care about it anymore. :)

primitives/src/balances.rs
clangenb: Thanks a lot, I think you are right, I have updated the documentation accordingly, I hope it is better now.

balances/src/lib.rs
clangenb: Changed the demurrage function itself to assume sensible defaults in case an overflow error happens.

PR #300, opened 2023-02-24
[balances] allow configurable source for the blocknumber

clangenb:
I consider this an intermediate step such that we can change the blocknumber source to the current parentchain blocknumber in the community sidechain. In the future, I think that we might want to abstract the whole `ApplyDemurrage` associated type. I will formulate an issue for this soon.

### DIFF
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -82,6 +82,8 @@ pub mod pallet {
 		type WeightInfo: WeightInfo;
 
 		type CeremonyMaster: EnsureOrigin<Self::Origin>;
+
+		type GetCurrentBlockNumber: GetCurrentBlockNumber<Self::BlockNumber>;
 	}
 
 	#[pallet::call]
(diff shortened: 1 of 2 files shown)


clangenb:
Not sure, if we merge this, I just found out that @echevrier has another workaround.

echevrier:
> Not sure, if we merge this, I just found out that @echevrier has another workaround.

My workaround is not working. So we have to check this one

brenzi:
This is one possible solution, the other is to just modify the demurrage rate such that we can use the sidechain block number (and change nothing in the code doing so)
Not sure which blocktime will be more steady in the long run, but demurrage doesn't need to be very precise

clangenb:
What I like about my approach is that it is more generic, and it allows arbitrary demurrage sources. I even considered making the demurrage function itself an associated type to simplify testing in some cases.

brenzi:
if we generalize, then we should generalize to the extent where even block timestamps can be used (demurrage is calculated based on time elapsed, not on the number of blocks.)
I see little benefit in merging this as is, but I can accept it

clangenb:
No this is fine, we don't have to merge it. I have been thinking myself that we should make it more generic. We could also pass the entire demurrage function as an associated type.

PR #309, opened 2023-05-03
mark rpcs as safe

pifragile:
* Closes https://github.com/encointer/encointer-node/issues/308#event-9053032926

### DIFF
--- a/bazaar/rpc/src/lib.rs
+++ b/bazaar/rpc/src/lib.rs
@@ -16,7 +16,6 @@
 
 use encointer_rpc::Error;
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
-use sc_rpc::DenyUnsafe;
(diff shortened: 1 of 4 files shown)


bazaar/rpc/src/lib.rs
```diff
@@ -56,6 +56,7 @@ where
 pub struct BazaarRpc<Client, Block, AccountId> {
 	client: Arc<Client>,
 	_marker: std::marker::PhantomData<(Block, AccountId)>,
+	#[allow(unused)]
 	deny_unsafe: DenyUnsafe,
```
clangenb, line 60: I doubt that we are ever going to use this flag again. I know this is a breaking API change, but I favour removing it now.

clangenb (commented):
Nice, as expected, but does anything speak against removing the `deny_unsafe` field?

clangenb (approved):
Nice, looks good to me!

PR #310, opened 2023-05-10
Faucet

### LINKED ISSUE #303: new pallet `encointer_reputation_commitments`
For the purpose of [personhood oracle](https://github.com/encointer/encointer-node/issues/300)

In order to get reputation linkability, we need an immutable record of commitments to purposes. While each purpose could maintain such registry for themselves, this is not possible for every purpose. Nostr. for example has no way of referring to a common consensus on commitments. Subsocial could, as they run their own parachain, but it would take integration effort which may slow down adoption. So the Encointer mainnet could provide such a simple registry:

```
pub fn commit_reputation(account_id. cid, cindex, purpose_id: u32, commitment: hash)
  if signature ok and not yet committed to *purpose*
    add commitment to registry double map (purpose, (account_id,cid,cindex))=commitment
pub fn register_purpose(description: String[24])
   append to array (descriptor)
   dispatch Event RegisteredCommitmentPurpose(purpose_id, descriptor)
```
where commitment is the hash of the requesting purpose account (i.e. nostr pubkey) and a random salt which should be unique per call and must later be provided to request a badge

the scope of purposes is global
including unit tests

### LINKED ISSUE #304: implement sybil-resistant faucet for KSM
As a first version of the faucet described in https://github.com/encointer/encointer-parachain/issues/172, we can implement the logic for our native token, the KSM

This avoids all complexity that comes with MultiAsset/MultiOrigin  XCM 

In order to "burn" used reputation, we probably should use this simple generic pallet in #303 where each instance of a faucet is a *purpose* which only allows a single use of a proof of attendance

pifragile:
* Closes #303 
* Closes #304


reputation-commitments/src/tests.rs
```diff
@@ -0,0 +1,399 @@
...
+			// commit without hash
+			// alice
+			assert_ok!(EncointerReputationCommitments::do_commit_reputation(
+				&alice.clone(),
+				cid,
+				11,
+				0,
+				None
+			));
+			assert_eq!(EncointerReputationCommitments::commitments((cid, 11), (0, &alice)), None);
```
brenzi, line 139: this doesn't test anything: if value is None, it is the same as if the key were not there at all. you can change alice to Bob and it will still pass here

reputation-commitments/src/tests.rs
```diff
@@ -0,0 +1,399 @@
...
+				0,
+				None
+			));
+			assert_eq!(EncointerReputationCommitments::commitments((cid, 11), (0, &bob)), None);
+			assert_eq!(
+				last_event::<TestRuntime>(),
+				Some(Event::CommitedReputation(cid, 11, 0, bob.clone(), None).into())
+			);
+
+			// same reputation, different purpose
```
brenzi, line 159: wrong comment. its same commitment, different proof of attendance

reputation-commitments/src/tests.rs
```diff
@@ -0,0 +1,399 @@
...
+
+		//ext.insert(community_identifiers(), vec![cid, cid2].encode());
+
+		ext.insert(participant_reputation((cid, 1), &alice), Reputation::VerifiedUnlinked.encode());
+		ext.insert(
+			participant_reputation((cid2, 1), &alice),
+			Reputation::VerifiedUnlinked.encode(),
+		);
+		ext.insert(participant_reputation((cid, 1), &bob), Reputation::VerifiedUnlinked.encode());
+		ext.insert(participant_reputation((cid2, 1), &bob), Reputation::VerifiedUnlinked.encode());
```
brenzi, line 261: cheeky Bob managed to be in two communities for the same cycle. Not relevant, but strange....

reputation-commitments/src/tests.rs
```diff
@@ -0,0 +1,399 @@
...
+			assert!(Commitments::<TestRuntime>::contains_key((cid2, 2), (0, &alice)));
+			assert!(Commitments::<TestRuntime>::contains_key((cid, 2), (0, &bob)));
+			assert!(Commitments::<TestRuntime>::contains_key((cid, 2), (1, &alice)));
+			assert!(Commitments::<TestRuntime>::contains_key((cid2, 2), (1, &bob)));
+
+			run_to_next_phase();
+			run_to_next_phase();
+			run_to_next_phase();
+
+			assert!(!Commitments::<TestRuntime>::contains_key((cid, 1), (0, &alice)));
```
brenzi, line 374: hmm, this actually means pruging did not work, right? this test seems to pass although purging failed

brenzi: I believe onCeremonyPhaseChange isn't even linked in the mock, so purging will never happen

scheduler/src/lib.rs
```diff
@@ -57,6 +57,8 @@ pub mod pallet {
 
 		/// Who to inform about ceremony phase change
 		type OnCeremonyPhaseChange: OnCeremonyPhaseChange;
+		type OnCeremonyPhaseChange2: OnCeremonyPhaseChange;
```
brenzi, line 60: that's not how to do it. you can link multiple callbacks by using tuples.
See: 
https://github.com/paritytech/polkadot/blob/bddd25c8fdabb8f5c540d21f70bf1af96809114e/runtime/common/src/traits.rs#L254-L255

faucet/src/tests.rs
```diff
@@ -0,0 +1,153 @@
...
+//
+// Encointer is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Unit tests for the encointer_balances module.
```
brenzi, line 17: pls adapt

faucet/src/tests.rs
```diff
@@ -0,0 +1,153 @@
...
+	communities::{Degree, Location},
+	reputation_commitments::{DescriptorType, FromStr},
+};
+use frame_support::{assert_err, assert_ok};
+use mock::{master, new_test_ext, RuntimeOrigin, TestRuntime};
+use sp_runtime::DispatchError;
+use test_utils::{helpers::*, storage::*, *};
+
+#[test]
+fn purpose_id_is_stored_correclty() {
```
brenzi, line 33: typo

brenzi: but: this test just tests the mock, not the dut. why is it here?
this test should be covered by the unit tests of reputation_commitments, right?

faucet/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
+	pub struct Pallet<T>(PhantomData<T>);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + encointer_reputation_commitments::Config {
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		type Currency: Currency<Self::AccountId>;
+		type FaucetMaster: EnsureOrigin<Self::RuntimeOrigin>;
```
brenzi, line 62: I believe the polkadot convention is to have `Origin` in the naming of origins. We may want to align this accross all pallets to just say `ControllerOrigin` ?

faucet/src/tests.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+#[test]
+fn minting_works() {
+	new_test_ext().execute_with(|| {
+		Balances::make_free_balance_be(&EncointerFaucet::account_id(), 101);
+		// free_balance - minimum_balance
+		assert_eq!(EncointerFaucet::pot(), 100);
+	});
+}
```
brenzi, line 79: misnomer: this test is not about minting, it tests if the `pot()` function returns the correct balance of the pot
and the test is brittle because it hardcodes the minimum_balance

faucet/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+	impl<T: Config> Pallet<T> {
+		/// The account ID of the faucet.
+		///
+		/// This actually does computation. If you need to keep using it, then make sure you cache the
+		/// value and only call this once.
+		pub fn account_id() -> T::AccountId {
+			T::PalletId::get().into_account_truncating()
+		}
+
```
brenzi, line 118: fine for now. Just a comment: looking ahead, I'd prefer if this would be called `pot_account(Option<FaucetIndex>)` and the account should be distinct per faucet instance - as there may be multiple faucets with different sponsors and dripping amounts or whitelisted cids for the same token

faucet/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+		fn build(&self) {
+			// Create faucet account
+			let account_id = <Pallet<T>>::account_id();
+			let min = T::Currency::minimum_balance();
+			if T::Currency::free_balance(&account_id) < min {
+				let _ = T::Currency::make_free_balance_be(&account_id, min);
+			}
+			<DripAmount<T>>::put(self.drip_amount);
```
brenzi, line 147: as DripAmount is in the genesis config, this should not be necessary

faucet/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+			<encointer_reputation_commitments::Pallet<T>>::do_register_purpose(
+				DescriptorType::from_str("EncointerFaucet").unwrap(),
+			)
+			.expect("In case of purpose registry overflow, we cannot use this pallet.");
```
brenzi, line 152: the problem with setting this up at genesis is that this will fail for runtime upgrades from earlier versions. So either we implement state migrations that initially link a purpose and a faucet or we find another solution.
As encointer has no root origin, we have no way of setting this link manually

brenzi: an alternative design approach would be:
the key for the StorageMap at `reputation_commitments::Commitments` can be either an integer like now or it is the account_id of the faucet and therefore derived automatically and uniquely and exclusively tied to a specific pot

```
enum PurposeIdType {
    external(u32),
    internal(AccountId),
}
```
this way, no registration is needed for internal/onchain use. you'd obviously need to adjust your check of a purpose exists because that ensure is only needed for external purposes

brenzi (changes requested):
looks very lean and clean, was a very pleasant review, thx

I do like logs for debugging, so I request that you print a log for every Event that you spawn. We do that in other pallets too and it really helps.

reputation-commitments/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+				cindex,
+				purpose,
+				account.clone(),
+				commitment_hash,
+			));
+			Ok(())
+		}
+
+		#[allow(deprecated)]
+		pub fn purge_registry(cindex: CeremonyIndexType) {
```
brenzi, line 124: please dispatch an Event here. helps debugging

brenzi:
for context. I have spawned off the next few issues to follow this one:
1. https://github.com/encointer/pallets/issues/312
2. https://github.com/encointer/encointer-parachain/issues/180
1. https://github.com/encointer/pallets/issues/311

reputation-commitments/src/tests.rs
pifragile: absolutely, good catch.

reputation-commitments/src/tests.rs
pifragile: it is linked in the mock. see [here](https://github.com/encointer/pallets/blob/7752ed427ad7b80437bc161a23a4840f6ae6f982/reputation-commitments/src/mock.rs#L52) and [here](https://github.com/encointer/pallets/blob/7752ed427ad7b80437bc161a23a4840f6ae6f982/test-utils/src/lib.rs#L256)

why would you assume that purging did not work? we assert that the key is not present anymore, so it was purged.

faucet/src/lib.rs
pifragile: hm, but this seems like a pattern we use everywhere, eg. [here](https://github.com/encointer/pallets/blob/90b70555507f83fb842a83dbbdee8c8877bf2899/communities/src/lib.rs#L447)

faucet/src/lib.rs
pifragile: i dont fully understand the issue with the genesis config. can you explain?

the second approach seems too specific to me, because we also need internal uses that have an u32 id, like for example in democracy.

reputation-commitments/src/lib.rs
pifragile: ok

faucet/src/lib.rs
pifragile: makes sense

faucet/src/lib.rs
brenzi: ok

brenzi (approved):

PR #319, opened 2023-06-01
Multiple faucets with whitelist and KSM deposit

### LINKED ISSUE #312: [faucet] allow multiple instances with individual community whitelists
starting from #304, allow anyone to instantiate a new faucet and define a whitelist with cid's that are eligit to drip
Still only for native token (KSM).

new requirements:
* unpermissioned faucet creation where a minimum of funds must be supplied upon creation
* when a faucet runs dry, it gets dissolved and remainders go to the treasury
* each faucet has a whitelist for cid's that are eligible. the whitelist can only be supplied upon creation
* root origin can dissolve a faucet and define a beneficiary accountid where all remaining funds are transferred

the creator of a faucet instance has no special rights whatsoever after creation.

pifragile:
* Closes #312


faucet/src/lib.rs
```diff
@@ -56,125 +63,277 @@ pub mod pallet {
...
+		pub fn close_faucet(
+			origin: OriginFor<T>,
+			faucet_account: T::AccountId,
+		) -> DispatchResultWithPostInfo {
+			let from = ensure_signed(origin)?;
+			let faucet = Self::faucets(&faucet_account).ok_or(<Error<T>>::InexsistentFaucet)?;
+
+			ensure!(from == faucet.creator, <Error<T>>::NotCreator);
+			ensure!(
+				<T as Config>::Currency::free_balance(&faucet_account) < faucet.drip_amount,
```
brenzi, line 226: I'm afraid this could go south: If faucet is created with an integer multiple of drip amount, there will be a moment when faucet balance = drip_amount. But you can't drip anymore because of keep_alive...it wont go below ED
to be safe, the test could be "if balance < 2*drip_amount

faucet/src/lib.rs
```diff
@@ -56,125 +63,277 @@ pub mod pallet {
...
+		ReserveIdentifierOf<T>: From<[u8; 8]>,
+	{
 		#[pallet::call_index(0)]
 		#[pallet::weight(10_000)]
+		pub fn create_faucet(
+			origin: OriginFor<T>,
+			name: FaucetNameType,
+			amount: BalanceOf<T>,
+			whitelist: WhiteListType,
+			drip_amount: BalanceOf<T>,
```
brenzi, line 93: I believe we lack plausibility tests here:
* ensure drip amount > ED

faucet/src/lib.rs
```diff
@@ -56,125 +63,277 @@ pub mod pallet {
...
+			let faucet_id_hash: H256 = T::Hashing::hash_of(&faucet_identifier).into();
+			let faucet_account = T::AccountId::decode(&mut faucet_id_hash.as_bytes())
+				.expect("32 bytes can always construct an AccountId32");
+
+			if <Faucets<T>>::contains_key(&faucet_account) {
+				return Err(<Error<T>>::FaucetAlreadyExists.into())
+			}
+
+			<T as Config>::Currency::reserve_named(
+				&Self::get_reserve_id(&faucet_account),
```
brenzi, line 117: It is dangerous if you start altering state before all verifications have been done:
If the caller has enough balance to reserve, but not enough for the initial amount specified, the funds will be reserved but the faucet won't be created and the reserved funds will be lost.

Or am I missing something?

faucet/src/lib.rs
```diff
@@ -56,125 +63,277 @@ pub mod pallet {
...
 			cid: CommunityIdentifier,
 			cindex: CeremonyIndexType,
 		) -> DispatchResultWithPostInfo {
 			let from = ensure_signed(origin)?;
 
+			if !<Faucets<T>>::contains_key(&faucet_account) {
+				return Err(<Error<T>>::InexsistentFaucet.into())
+			}
+
+			let faucet = Self::faucets(&faucet_account).ok_or(<Error<T>>::InexsistentFaucet)?;
```
brenzi, line 155: seems redundant why test first instead of fetching and catching the possible error?

faucet/src/tests.rs
```diff
@@ -17,138 +17,644 @@
...
+		assert_err!(
+			EncointerFaucet::create_faucet(
+				RuntimeOrigin::signed(alice.clone()),
+				FaucetNameType::from_str("Some Faucet Name").unwrap(),
+				100,
+				whitelist_input.clone(),
+				10
 			),
-			DispatchError::BadOrigin,
+			Error::<TestRuntime>::InsuffiecientBalance
```
brenzi, line 112: If my comment above was right, faucet creation fails, but the reserve amount is locked anyway. should be tested here and fixed if  necessary

brenzi (commented):
looks very nice overall

faucet/src/lib.rs
pifragile: @clangenb confirmed to me that extrinsics are now atomic, so if anything fails in the extrinsic, the rest will be rolled back.

brenzi: a major advantage over solidity smart contracts

brenzi (approved):

PR #321, opened 2023-06-01
Introduce `Business` type containing controller account and the `BusinessData`

### LINKED ISSUE #318: How should we get the controller account of a business?
We remove the controller account information while getting the businesses in the `encointer_bazaarGetBusinesses` RPC. How is the app supposed to know, which offerings belong to whihcch business?

clangenb:
* Closes #318 

The nice thing is that this is a client update only, so we can deploy this fix without doing a runtime upgrade.


pifragile (approved):
lgtm

PR #326, opened 2023-06-07
faucet benchmarks

pifragile:


pifragile:
pls wait with merging, more to come

brenzi:
Am I wrong or do you only benchmark with happy inputs?
What if the description or whitelist have max length? If we simplify, we should benchmark the worst case only

brenzi (approved):

PR #332, opened 2023-07-22
upgrade to polkadot-v1.0.0

brenzi:
noteworthy causes for change:
* no more `T::BlockNumber` https://github.com/paritytech/substrate/pull/14437
* deprecated GenesisBuild: https://github.com/paritytech/substrate/pull/14306
* https://github.com/paritytech/substrate/pull/14437
* https://github.com/paritytech/substrate/pull/14290
* https://github.com/paritytech/substrate/pull/14261

once merged, reset branch polkadot-v1.0.0

### DIFF
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -25,7 +25,7 @@ use encointer_primitives::{
 use frame_support::{
 	dispatch::DispatchResult,
 	ensure,
-	traits::{tokens::fungibles, Get},
+	traits::{tokens::fungibles, Get, GenesisBuild},
 };
 use frame_system::{self as frame_system, ensure_signed};
 use log::{debug, info};
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -25,9 +25,9 @@ use encointer_primitives::{
 use frame_support::{
 	dispatch::DispatchResult,
 	ensure,
-	traits::{tokens::fungibles, Get, GenesisBuild},
+	traits::{tokens::fungibles, Get},
 };
-use frame_system::{self as frame_system, ensure_signed};
+use frame_system::{self as frame_system, ensure_signed, pallet_prelude::BlockNumberFor};
 use log::{debug, info};
 use sp_runtime::DispatchError;
 use sp_std::convert::TryInto;
@@ -126,8 +126,9 @@ pub mod pallet {
 		}
 	}
 
+	#[derive(frame_support::DefaultNoBound)]
 	#[pallet::genesis_config]
-	pub struct GenesisConfig {
+	pub struct GenesisConfig<T> {
 		/// Example parameter tuning for the fee-conversion
 		///
 		/// [CC]:	1 Unit of Community Currency
@@ -142,18 +143,12 @@ pub mod pallet {
 		///
 		/// FCF = 0.01 / (20 * 5*10^-12 [KKSM]) = 0.01 / (20 * 5 * 10e-12) = 100_000
 		pub fee_conversion_factor: FeeConversionFactorType,
-	}
-
-	#[cfg(feature = "std")]
-	#[allow(clippy::derivable_impls)]
-	impl Default for GenesisConfig {
-		fn default() -> Self {
-			Self { fee_conversion_factor: Default::default() }
-		}
+		#[serde(skip)]
+		pub _config: sp_std::marker::PhantomData<T>,
 	}
 
 	#[pallet::genesis_build]
-	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+	impl<T: Config> BuildGenesisConfig for GenesisConfig<T> {
 		fn build(&self) {
 			<FeeConversionFactor<T>>::put(self.fee_conversion_factor);
 		}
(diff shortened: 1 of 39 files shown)


brenzi:
still getting error logs during tests, deemed insubstantial:
`Jul 22 16:34:17.299 ERROR runtime::system: Logic error: Account already dead when reducing provider`

brenzi:
hmm, this may be relevant too, even if older: https://github.com/paritytech/substrate/pull/14261/files

ceremonies/src/lib.rs
```diff
@@ -1032,6 +1032,7 @@ pub mod pallet {
 	#[pallet::getter(fn meetup_time_offset)]
 	pub(super) type MeetupTimeOffset<T: Config> = StorageValue<_, MeetupTimeOffsetType, ValueQuery>;
 
+	#[derive(frame_support::DefaultNoBound)]
```
clangenb, line 1035: Without double-checking what this is about, I can imagine that this does only need to be derived in `std` and if you derive it in no-std, you get missing serialize traitbound error in the node.

communities/src/lib.rs
```diff
@@ -467,22 +468,17 @@ pub mod pallet {
 	#[pallet::getter(fn max_speed_mps)]
 	pub(super) type MaxSpeedMps<T: Config> = StorageValue<_, MaxSpeedMpsType, ValueQuery>;
 
+	#[derive(frame_support::DefaultNoBound)]
 	#[pallet::genesis_config]
```
clangenb, line 472: dito

faucet/src/lib.rs
```diff
@@ -282,20 +282,17 @@ pub mod pallet {
 			reserve_id.into()
 		}
 	}
+
+	#[derive(frame_support::DefaultNoBound)]
```
clangenb, line 286: dito

reputation-commitments/Cargo.toml
```diff
@@ -46,7 +46,6 @@ std = [
     "sp-std/std",
     "sp-runtime/std",
     "sp-core/std",
-    "sp-runtime/std",
```
clangenb, line 49: why did you remove this std?

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2023-01-13" 
+channel = "nightly-2023-05-22"
```
clangenb, line 2: This is still quite far in the past, any reason for that?

balances/src/lib.rs
```diff
@@ -126,8 +126,9 @@ pub mod pallet {
 		}
 	}
 
+	#[derive(frame_support::DefaultNoBound)]
```
clangenb, line 129: dito

clangenb (commented):
looks good in general, some minor questions, and I might have spotted the missing serialize implementation error.

clangenb:
> Jul 22 16:34:17.299 ERROR runtime::system: Logic error: Account already dead when reducing provider
I made an issue of this. This is suspicious, in the worst case, it could mean that we forget to increment a provider count when we should, and as a result we might kill an account, which should not be killed yet. #333

reputation-commitments/Cargo.toml
brenzi: duplicate of two lines above

rust-toolchain.toml
brenzi: yeah, it's the reference bightly parity mentions in the polkadot release

ceremonies/src/lib.rs
brenzi: hmmm, this can be found all over the place in frame
https://github.com/paritytech/substrate/blob/b8d485fd4942539c438bf25a7823e410360fa053/frame/grandpa/src/lib.rs#L337

brenzi: the problem is not restricted to no_std. It fails hwen I do: `SKIP_WASM_BUILD=1 cargo build --release`

clangenb: Yes, I was wrong.

clangenb:
> hmm, this may be relevant too, even if older: [paritytech/substrate#14261](https://github.com/paritytech/substrate/pull/14261)

Hmm, this could mean that we need to enable the `serde_derive` feature here: https://github.com/encointer/encointer-node/blob/0e4b92736783483ec35d21ba09fc0d20fdad9346/runtime/Cargo.toml#L21, which we currently only do in no-std.

Unfortunately, I just tried it and it seems to introduce an std-leackage....

clangenb:
> > hmm, this may be relevant too, even if older: [paritytech/substrate#14261](https://github.com/paritytech/substrate/pull/14261)
> 
> Hmm, this probably means that we need to enable the `serde_derive` feature here: https://github.com/encointer/encointer-node/blob/0e4b92736783483ec35d21ba09fc0d20fdad9346/runtime/Cargo.toml#L21, which we currently only do in std.
> 
> Unfortunately, I just tried it and it seems to introduce an std-leackage....

I should have checked it more thoroughly, it does fix the runtime build. It is just the node, which also needs some fixes then. However, if the node requires this flag now in any case, it is better to enable encointer-primitive's `serde_derive` flag from within the pallets, such that the node does not need to care about it.

clangenb (approved):
LGTM

PR #336, opened 2023-07-26
added fields AnnouncementSigner and CommunityRules to CommunityMetadata

### LINKED ISSUE #334: introduce per-community announcement signer
As the Encointer Wallet app features community leadership announcements for enhanced coordination, community metadata should feature an authorized account which can author announcements to be displayed in the app.
No further powers are assigned to this account.

as community metadata is poised to be subject to community democracy, the community can vote on who has the right to author announcements

While this introduces some kind of representational centralization, experience has shown that bootstrapping a community without a dedicated leadership guidance is not realistic

As announcements are entirely offchain and possibly published with nostr, we should not assume a specific crypto for signing other than the pubkey length. Nostr uses [schnorr signatures on secp256k1](https://bips.xyz/340) which also features 32-byte pubkeys.
In the best case, we could extend [MultiSigner](https://paritytech.github.io/substrate/master/sp_runtime/enum.MultiSigner.html) to feature [BIP340](https://bips.xyz/340), but we may need to use our own wrapper for now, which is future-proof

```
pub enum AnnouncementSigner {
   substrate(MultiSigner),
   bip340([u8; 32])
}
```

### LINKED ISSUE #335: introduce `CommunityRules` enum in community metadata
Currently, the only community type on encointer are *LoCo* communities.
With BeeDance, we introduce an event-based type which has different rules and there may even be an intermediate type with rule relaxations 

To keep it easy and flexible, an enum `CommunityRules`shall start with:
* LoCo (default)
* LoCoLight (UI enforces relaxed rules but same onchain protocol)
* BeeDance

| type | globally synched cycle day | globally synched cycle time of day | globally unique min-dist locations | demurrage |
|-----|-------|-------|--------|----------|
| LoCo | X | X | >100m distance | X |
| LoCoLight (less friction but less secure and less decentralized)| X | can be defined by community leads on day ahead | coarsely only (city level).  precice locations can be defined by community leadership and can be very close | X |
| BeeDance | - | - | - | - |

brenzi:
closes #334 
closes #335 

breaking change! 
includes a storage migration to v2 for the communities pallet

### DIFF
--- a/communities/src/migrations.rs
+++ b/communities/src/migrations.rs
@@ -262,7 +262,7 @@ pub mod v2 {
 	use super::*;
 	use crate::migrations::{v0::UnboundedCommunityMetadata, v1::CommunityMetadataV1};
 	use encointer_primitives::{
-		common::FromStr,
+		common::{BoundedIpfsCid, FromCropping, FromStr},
 		communities::{AnnouncementSigner, CommunityRules},
 	};
 	use scale_info::named_type_params;
@@ -272,7 +272,7 @@ pub mod v2 {
 	impl<T: Config + frame_system::Config> OnRuntimeUpgrade for Migration<T> {
 		#[cfg(feature = "try-runtime")]
 		fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
-			assert_eq!(StorageVersion::get::<Pallet<T>>(), 0, "can only upgrade from version 0");
+			assert!(StorageVersion::get::<Pallet<T>>() < 2, "can only upgrade from version 0 or 1");
 
 			// TODO
 			// For community metadata, we do not need any checks, because the data is bounded already due to the CommmunityMetadata validate() function.
@@ -282,11 +282,11 @@ pub mod v2 {
 
 		fn on_runtime_upgrade() -> Weight {
 			let weight = T::DbWeight::get().reads(1);
-			if StorageVersion::get::<Pallet<T>>() != 0 {
+			if StorageVersion::get::<Pallet<T>>() >= 2 {
 				log::warn!(
 					target: TARGET,
 					"skipping on_runtime_upgrade: executed on wrong storage version.\
-				Expected version 0"
+				Expected version 0 or 1"
 				);
 				return weight
 			}
@@ -295,33 +295,25 @@ pub mod v2 {
 				|k: CommunityIdentifier, meta: UnboundedCommunityMetadata| {
 					info!(target: TARGET, "     Migrating community metadata for {:?}...", k);
 
-					// here we just transcode because it seems that the storage representation of Vec and BoundedVec is the same.
-					// as long as we check the bounds in pre_upgrade, we should be fine.
-					let meta_v1 = CommunityMetadataV1::decode(&mut meta.encode().as_ref())
-						.unwrap_or_default();
-
-					Some(
-						CommunityMetadataType::new(
-							meta_v1.name,
-							meta_v1.symbol,
-							meta_v1.assets,
-							meta_v1.theme,
-							meta_v1.url,
-							Some(AnnouncementSigner::default()),
-							CommunityRules::default(),
-						)
-						.unwrap_or_default(),
-					)
+					Some(CommunityMetadataType {
+						name: PalletString::from_cropping(meta.name.into()),
+						symbol: PalletString::from_cropping(meta.symbol.into()),
+						assets: BoundedIpfsCid::from_cropping(meta.assets.into()),
+						theme: meta.theme.map(|theme| BoundedIpfsCid::from_cropping(theme.into())),
+						url: meta.url.map(|url| PalletString::from_cropping(url.into())),
+						announcement_signer: Some(AnnouncementSigner::default()),
+						rules: CommunityRules::default(),
+					})
 				},
 			);
 
-			StorageVersion::new(1).put::<Pallet<T>>();
+			StorageVersion::new(2).put::<Pallet<T>>();
 			weight.saturating_add(T::DbWeight::get().reads_writes(1, 2))
 		}
 
 		#[cfg(feature = "try-runtime")]
 		fn post_upgrade(state: Vec<u8>) -> Result<(), &'static str> {
-			assert_eq!(StorageVersion::get::<Pallet<T>>(), 1, "must upgrade");
+			assert_eq!(StorageVersion::get::<Pallet<T>>(), 2, "must upgrade");
 			//TODO
 			Ok(())
 		}
--- a/communities/src/migrations.rs
+++ b/communities/src/migrations.rs
@@ -119,7 +119,7 @@ mod v0 {
 
 pub mod v1 {
 	use super::*;
-	use encointer_primitives::common::BoundedIpfsCid;
+	use encointer_primitives::{common::BoundedIpfsCid, communities::CommunityRules};
 
 	#[derive(
 		Encode, Decode, Default, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen,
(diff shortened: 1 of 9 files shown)


brenzi:
caveat: the rpc reads CommunityMetadata.name and will break if this can't be decoded. therefore we need state migration (manual or automated)

communities/src/migrations.rs
```diff
@@ -103,43 +220,70 @@ pub mod v1 {
...
+				return T::DbWeight::get().reads(1)
 			}
 
-			// we do not actually migrate any data, because it seems that the storage representation of Vec and BoundedVec is the same.
-			// as long as we check the bounds in pre_upgrade, we should be fine.
+			CommunityMetadata::<T>::translate::<UnboundedCommunityMetadata, _>(
+				|k: CommunityIdentifier, meta: UnboundedCommunityMetadata| {
+					info!(target: TARGET, "     Migrating community metadata for {:?}...", k);
+					translated.saturating_inc();
+					Some(meta.migrate_to_v2())
```
pifragile, line 260: ok, you are saying that this works for `v0` -> `v2` as well as `v1` -> `v2` because the storage representation of `UnboundedCommunityMetadata` and `CommunityMetadataV1` is the same?

communities/src/migrations.rs
```diff
@@ -181,7 +325,8 @@ pub mod v1 {
 #[cfg(feature = "try-runtime")]
 mod test {
```
pifragile, line 326: tests now only cover the scenarion 0 -> 2, would be good to also have 1-> 2

communities/src/migrations.rs
pifragile: actually pretty sure it wont work for `v1` -> `v2` because `from_cropping` takes `Vec<u8>` and this will be the wrong type if the input is a `CommunityMetadataV1`.

communities/src/migrations.rs
brenzi: agreed

communities/src/migrations.rs
brenzi: generic didn't work immediately, but actually not necessary. tested both cases now

communities/src/migrations.rs
brenzi: test added

communities/src/migrations.rs
pifragile: hmm ok. it seems counterintuitive to me but it probably works, because we somehow implicitly "cast" a`CommunityMetadataV1` into a `UnboundedCommunityMetadata`, because we operate direclty on the storage value..

brenzi:
while parity tends to do single-step migrations, I believe this case here is much simpler to understand and comes with less boilerplate if we allow migration from v0->v2 and v1->v2 in a single move. v0->v1 was a noop with bound checks only

primitives/src/common.rs
```diff
@@ -14,28 +14,20 @@
...
 use sp_std::vec::Vec;
 
 /// Substrate runtimes provide no string type. Hence, for arbitrary data of varying length the
 /// `Vec<u8>` is used. In the polkadot-js the typedef `Text` is used to automatically
 /// utf8 decode bytes into a string.
-#[cfg(not(feature = "std"))]
-pub type UnboundedPalletString = Vec<u8>;
-
-#[cfg(feature = "std")]
-pub type UnboundedPalletString = String;
```
clangenb, line 38: I like that the old types live only in migrations file now 👍

primitives/src/common.rs
```diff
@@ -51,20 +43,22 @@ impl FromStr for PalletString {
...
-	#[cfg(feature = "std")]
-	fn as_bytes_or_noop(&self) -> &[u8] {
-		self.as_bytes()
+impl FromCropping for PalletString {
+	fn from_cropping(data: Vec<u8>) -> Self {
+		if data.len() <= Self::bound() {
+			Self::try_from(data).expect("can't be too long")
+		} else {
+			Self::try_from(data[..Self::bound()].to_vec()).expect("can't be too long now")
+		}
```
clangenb, line 56: For bounded vec they use the term `truncating_from`, I suggest we use the same: https://paritytech.github.io/substrate/master/frame_support/storage/bounded_vec/struct.BoundedVec.html#method.truncate_from

primitives/src/communities.rs
```diff
@@ -229,20 +230,46 @@ impl Location {
...
+pub enum AnnouncementSigner {
+	Substrate(MultiSigner),
+	Bip340([u8; 32]),
+}
+
+impl Default for AnnouncementSigner {
+	fn default() -> Self {
+		AnnouncementSigner::Bip340([0u8; 32])
+	}
+}
```
clangenb, line 245: Does it make sense to have a default here? Or is this because the storage is still a value query?

primitives/src/communities.rs
```diff
@@ -229,20 +230,46 @@ impl Location {
...
+impl MaxEncodedLen for AnnouncementSigner {
+	fn max_encoded_len() -> usize {
+		34
+	}
+}
```
clangenb, line 250: It should not have problems deriving that.

primitives/src/communities.rs
clangenb: I see it is only an option in the metadata. So I have to ask again. Why are we deriving a default?

clangenb (commented):
Looks good to me, only some minor comments. I like that we are no longer afraid of storage migrations. XD

primitives/src/communities.rs
brenzi: Not necessary, agreed

primitives/src/communities.rs
brenzi: The problem is that MultiSigner doesn't implement MaxEncodedLen it seems

clangenb: Hmm, should we maybe just go for an account id anyhow, what is the benefit of having the multisigner?

primitives/src/common.rs
brenzi: Haha, was looking for that but didn't find it in the docs....

primitives/src/communities.rs
brenzi: removed it altogether

primitives/src/communities.rs
```diff
@@ -229,20 +229,36 @@ impl Location {
...
-	pub symbol: UnboundedPalletString,
-	/// IPFS cid to assets necessary for community branding
-	pub assets: IpfsCid,
-	/// ipfs cid for style resources
-	pub theme: Option<IpfsCid>,
-	/// optional link to a community site
-	pub url: Option<UnboundedPalletString>,
+pub enum AnnouncementSigner {
+	Bip340([u8; 32]),
+}
```
clangenb, line 239: If we'd use the Account Id here, we'd get base58 encoding in the common UIs, this will be reflected as a hex. I am uncertain if this is what we want, but I wanted to mention this.

brenzi: this will be a nostr pubkey and needs no substrate UI encoding

brenzi: and accountid doesn't support bip340

clangenb: I agree with the first, so absolutely fine with me, but yeah I was rather thinking of the `AccountId32`, which is agnostic to the sig-scheme not the  `T::AccountId`, which has to satisfy some trait bounds. However, I agree that we can leave it like that. In the future, we might want to have a:

```
struct Bip340/NostrPub([u8; 32]); 
```

to implement some methods on it, but this is a non-breaking change in terms of encoding and can also wait.

clangenb (approved):
All questions satisfied! :) Looks good to me

PR #337, opened 2023-07-26
add CommunityRules enum to CommunityMetadata

### LINKED ISSUE #335: introduce `CommunityRules` enum in community metadata

brenzi:
closes #335

### DIFF
--- a/primitives/src/communities.rs
+++ b/primitives/src/communities.rs
@@ -230,18 +230,18 @@ impl Location {
 }
 
 #[derive(
-Encode,
-Decode,
-Copy,
-Clone,
-PartialEq,
-Eq,
-Default,
-RuntimeDebug,
-PartialOrd,
-Ord,
-TypeInfo,
-MaxEncodedLen,
+	Encode,
+	Decode,
+	Copy,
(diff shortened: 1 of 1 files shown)


brenzi:
caveat: the rpc reads CommunityMetadata.name and will break if this can't be decoded. therefore we need state migration (manual or automated)

PR #345, opened 2023-08-21
fix sufficients error

### LINKED ISSUE #333: Bug: reducing provider references when account is already dead.
In the worst case, it could mean that we forget to increment a provider count when we should, and as a result we might kill an account, which should not be killed yet.

brenzi:
closes #333 

* added a unit test that reproduces the issue
* error was: issue() didn't check for account creation to inc_sufficients

consequence:
current mainnet state is dirty. we have two options
1. runtime state migration checking minimal balances on all accounts and recalculating sufficients and providers. Risk: overloading the block
2. lazy dispatchable `check_sufficients` which does the above per account. combined with a script that calls this extrinsic for every account: Risk: none

### DIFF
--- a/balances/src/lib.rs
+++ b/balances/src/lib.rs
@@ -322,8 +322,8 @@ impl<T: Config> Pallet<T> {
 		who: &T::AccountId,
 		amount: BalanceType,
 	) -> DispatchResult {
-		if !Balance::<T>::contains_key(community_id, &who) {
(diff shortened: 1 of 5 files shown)


clangenb (approved):
Good catch. Yeah, this is indeed bad that the sufficients count is wrong on the main net.

I don't see a third option either, and would probably go for the lazy approach.

PR #346, opened 2023-08-21
fix assignment bug

pifragile:
* Closes https://github.com/encointer/encointer-wallet-flutter/issues/1459

### DIFF
--- a/ceremonies/assignment/src/lib.rs
+++ b/ceremonies/assignment/src/lib.rs
@@ -103,7 +103,9 @@ pub fn assignment_fn_inverse(
 	let mut max_index = assignment_params.m.saturating_sub(meetup_index) / assignment_count;
 	let mut result: Vec<ParticipantIndexType> = Vec::with_capacity(max_index as usize);
 	// ceil
-	if (assignment_params.m as i64 - meetup_index as i64).rem_euclid(assignment_count as i64) != 0 {
+	if (assignment_params.m.saturating_sub(meetup_index) as i64).rem_euclid(assignment_count as i64) !=
+		0
+	{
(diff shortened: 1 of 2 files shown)


ceremonies/assignment/src/lib.rs
```diff
@@ -103,7 +103,9 @@ pub fn assignment_fn_inverse(
 	let mut max_index = assignment_params.m.saturating_sub(meetup_index) / assignment_count;
 	let mut result: Vec<ParticipantIndexType> = Vec::with_capacity(max_index as usize);
 	// ceil
-	if (assignment_params.m as i64 - meetup_index as i64).rem_euclid(assignment_count as i64) != 0 {
+	if (assignment_params.m.saturating_sub(meetup_index) as i64).rem_euclid(assignment_count as i64) !=
+		0
+	{
```
clangenb, line 108: Nice, catch. Do you have an idea, why this was noticed so late?

ceremonies/src/tests.rs
```diff
@@ -3234,3 +3234,80 @@ fn is_endorsed_works() {
 		assert_eq!(EncointerCeremonies::is_endorsed(&alice, &(cid, 9)), None);
 	});
 }
+
+#[test]
+fn participants_assigned_mateches_participants_registered() {
```
clangenb, line 3239: There is a typo.

And please link the issue that led to the introduction of this test.

clangenb (commented):
Awesome catch, I only noted a cosmetic issue.

clangenb (approved):
Nice, thanks for the fix. LGTM!

Issue #347, opened 2023-09-06
Democracy Future Work Tracking Issue


pifragile:
This is a list of features that should/could be implemented in the future of the democracy pallet:

- Decide which matters can be voted on and implement
- Use timestamps instead of blocks to be consistent with cycles
- Use reputation commitment pallet for proposals
- Pass proposal enactment errors to user
- Lazy proposal update
- Persist electorate size in proposal

PR #348, opened 2023-09-14
democracy benchmarks

pifragile:
* Pallet side of https://github.com/encointer/encointer-node/issues/347


democracy/src/benchmarking.rs
```diff
@@ -0,0 +1,91 @@
...
+		));
+
+		frame_support::storage::unhashed::put_raw(&participant_reputation((cid, 3), &zoran), &Reputation::VerifiedUnlinked.encode());
+		frame_support::storage::unhashed::put_raw(&participant_reputation((cid, 4), &zoran), &Reputation::VerifiedUnlinked.encode());
+		frame_support::storage::unhashed::put_raw(&participant_reputation((cid, 5), &zoran), &Reputation::VerifiedUnlinked.encode());
+		let reputation_vec: ReputationVecOf<T> = BoundedVec::try_from(vec![
+			(cid, 3),
+			(cid, 4),
+			(cid, 5),
+		]).unwrap();
```
brenzi, line 54: wouldn't the worst case complexity be 5/5 reputation?

democracy/src/benchmarking.rs
```diff
@@ -0,0 +1,91 @@
...
+		frame_system::Pallet::<T>::set_block_number(
+			frame_system::Pallet::<T>::block_number() + 1u32.into(),
+		);
+		frame_system::Pallet::<T>::on_initialize(frame_system::Pallet::<T>::block_number());
+	}
+}
+
+benchmarks! {
+
+	submit_proposal {
```
brenzi, line 26: Is the number of active proposals bounded?
It may make sense to restrict the number of proposals per reputable to one per cycle. but that can happen in another PR

brenzi (approved):
LGTM

democracy/src/benchmarking.rs
pifragile: no, because we only allow voting with reputations in the following range of cindexes: [proposal_start - reputation_life_time + proposal_lifetime, proposal_start - 2], because those are the reputations guaranteed not to change during the lifetime of a proposal.

democracy/src/benchmarking.rs
pifragile: no it is not bounded. i added this to https://github.com/encointer/pallets/issues/347

pifragile:
i cannot merge even if approved, so pls do merge :)

PR #353, opened 2023-09-21
use the unsigned fixed type for balances: `U64F64`

clangenb:


communities/src/lib.rs
```diff
@@ -106,9 +105,6 @@ pub mod pallet {
 			community_metadata
 				.validate()
 				.map_err(|_| <Error<T>>::InvalidCommunityMetadata)?;
-			if let Some(i) = nominal_income {
-				validate_nominal_income(&i).map_err(|_| <Error<T>>::InvalidNominalIncome)?;
-			}
```
clangenb, line 111: Validate only validated if the value is negative; type safety gives us this guarantee now already.

primitives/src/balances.rs
```diff
@@ -137,11 +137,14 @@ pub fn demurrage_factor(demurrage_per_block: Demurrage, elapsed_blocks: u32) ->
...
+		None => return 0u64.into(),
 	};
 
 	// exponent <= 0; hence return value [0, 1)
-	exp(exponent).unwrap_or_else(|_| 0.into())
+	let f: I64F64 = exp(exponent).unwrap_or_else(|_| 0.into());
+
+	// Safe conversion. The result of an exponential function can't be negative.
+	BalanceType::from_le_bytes(f.to_le_bytes())
 }
```
clangenb, line 148: I don't really see a reason not to trust this, do you?

primitives/src/balances.rs
```diff
@@ -215,26 +218,26 @@ mod tests {
 
 	#[test]
 	fn demurrage_works() {
-		let bal = BalanceEntry::<u32>::new(1.into(), 0);
+		let bal = BalanceEntry::<u32>::new(1u32.into(), 0);
```
clangenb, line 221: Implicit type is i32, which can obviously not be `into`-ed.

primitives/src/balances.rs
```diff
@@ -195,6 +204,15 @@ impl Convert<u128, BalanceType> for EncointerBalanceConverter {
 	}
 }
 
+#[allow(non_snake_case)]
+pub fn I64F64_to_U64F64(source: I64F64) -> Option<U64F64> {
+	if source.is_negative() {
+		return None
+	}
+
+	Some(U64F64::from_le_bytes(source.to_le_bytes()))
```
brenzi, line 213: I don't like this because byte order is not what makes this work and it confused me. what you do could be done with scale encode and decode and it would be at least obvious that it is a hacky typecast

brenzi (changes requested):
very nice cleanup. only one hack I don't like ;-)

brenzi (approved):
LGTM

primitives/src/balances.rs
clangenb: I agree that the LE is confusing, I have made it more trivial and just use `to_bits` and `from_bits` (much better) and also made a remark about the conversion/

I don't think it is good to to use the scale codec because it is much more inefficient than using the `#[inline]` bit conversions that operate on a stack instead of on the heap.

With the `from_bits` and `to_bits` conversions, I don't even consider this hacky anymore.

PR #355, opened 2023-09-27
vouches pallet MVP

### LINKED ISSUE #354: EncointerVouches MVP
time for a new pallet: encointer_vouches

the need for extreme social scalability with minimal friction led us here. therefore, for this product prioritize usability over other concerns.

See also: https://forum.encointer.org/t/scaling-encointer-fast-and-globally-through-staked-social-vouches/95

## spec

Basically, anyone or anything (identified by an account) should be able to make a statement about anyone or anything (identified by an account). as broad as that. statements can be

**Timestamp**
Every vouch has a timestamp and we may introduce a lazy garbage collection after a globally defined TTL

**VouchTypes** (statement from the perspective of the sender)
* encountered human (physical or virtual)
* encountered object (physical or virtual)
* visited event (physical or virtual)
* visited place (physical or virtual)

**VouchQuality**
* Rating(u8) _only for non-human VouchType_
* Badge(ipfs hash)

**Protection**
* Perishable _default_
* Protected _subject to attestee placing a deposit_

#### V1 MVP

* `vouch_for(AccountId, VouchType, VouchQuality)
`
#### V2

* ` restrict_vouches(Vec<AccountId>) `_restrict who may vouch for you (i.e. only registered event participants)_
* `restrict_qualities(Vec<VouchQuality>)` _whitelist ratings or specific badges and reject all others_
* `restrict_type(VouchType)` _deny all other types_
* `purge_outdated_vouch(attestee, attester)` _lazy cleanup_
* `protect_vouches()` _if the senderwants to keep their vouches and prevent purging, they can place a deposit per vouch _
* `unprotect_vouches()` _lift a protection and get back the deposit_

#### PoA

Event Organizers could present a QR code to their attendees which they scan and then send an extrinsic vouching for that event: `vouchFor(eventAccountId, visitedEvent, _)`

_caveat: even people who are not present could observe onchain activity and vouch for the event they don't attend_

To increase trust in this PoA, the event organizer could offer the service of scanning all attendees and vouching for their attendance: `vouchFor(attendeeAccountId, encounteredHuman, _)`

_caveat: event organizers have absolute power over their PoA's and could discriminate_

#### Treasure Hunts / Geocashing

Increased gamification: for a treasure hunt, people need to find QR codes either in physical locations or virtual spaces. Sponsors could issue NFT's or other reward tokens to everyone who found the treasure
`vouchFor(objectAccountId, encounteredObject, _)`

#### Decentralized LinkedIn

people could attest certain qualities of others in the form of awarding them badges along with their vouch. Badges can have metadata and images (the json format for badges has yet to be specified):
* "Knows Stuff"
* "Was at College with me"
* "Morally Trustworthy"

caveats: 
* weak privacy, everything public
* potential for bullying. we should uphold a social contract that badges may strictly be positive statements

#### Decentralized TripAdvisor

Restaurants or Hotels can display their QR code and guests can place a rating and optionally some badges like 
* "family friendly"
* "dog friendly"

brenzi:
closes #354


vouches/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+	pub enum Error<T> {
+		/// The calling attester has too many vouches for this attestee
+		TooManyVouchesForAttestee,
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn vouches)]
+	pub(super) type Vouches<T: Config> = StorageDoubleMap<
+		_,
+		Identity,
```
brenzi, line 98: I'm undecided on the key derivation. we may just go the same Blake2_128Concat like the system pallet

primitives/src/vouches.rs
```diff
@@ -0,0 +1,64 @@
...
+	Virtual,
+	Physical,
+}
+
+#[derive(Default, Encode, Decode, Copy, Clone, PartialEq, Eq, Debug, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "serde_derive", derive(Serialize, Deserialize))]
+#[repr(u8)]
+pub enum VouchType {
+	#[default]
+	Unspecified = 0,
```
brenzi, line 38: Here I'm not sure if we'd rather want to drop the unspecified field and instead have an Option<VouchType> field in Vouch. I chose to do it like this so people make an explicit choice in generic UI's like js/apps

clangenb: I agree, it is always a hassle to interpret the ambiguity of None vs key inexistent in the clients.

primitives/src/vouches.rs
```diff
@@ -0,0 +1,64 @@
...
+	KnownHuman = 10,
+	EncounteredHuman(PresenceType) = 20,
+	EncounteredObject(PresenceType) = 30,
+	VisitedEvent(PresenceType) = 40,
+	VisitedPlace(PresenceType) = 50,
```
clangenb, line 43: What do you want to achieve with the enum values and the `repr(u8)`? For backwards compatibility, we would always want to append at the end of the enum.

vouches/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
+name = "pallet-encointer-vouches"
+version = "1.2.0"
+authors = ["encointer.org <alain@encointer.org> and Parity Technologies <admin@parity.io>"]
```
clangenb, line 4: Is parity technologies here with the outlook of being included in the fellowship repo?

vouches/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+use encointer_primitives::vouches::{Vouch, VouchQuality, VouchType};
+use frame_system::{self as frame_system, ensure_signed, pallet_prelude::OriginFor};
+use log::info;
+pub use pallet::*;
+use sp_std::convert::TryInto;
+pub use weights::WeightInfo;
+// Logger target
+const LOG: &str = "encointer::vouches";
+
+#[cfg(feature = "runtime-benchmarks")]
```
clangenb, line 29: I would like this to be `cfg(any(test, feature = "runtime-benchmarks"))` then we can at least test if the benchmarks work in std.

vouches/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+    "codec/std",
+    "log/std",
+    "scale-info/std",
+    "approx/std",
+    # local deps
+    "encointer-primitives/std",
+    # substrate deps
+    "frame-support/std",
+    "frame-system/std",
+    "sp-std/std",
```
clangenb, line 42: I was wondering why this works, but I see now that you don't compile the benchmarks for this pallet with the tests. I think we need a `frame-benchmarking?/std` here.

vouches/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use core::marker::PhantomData;
+use encointer_primitives::vouches::{Vouch, VouchQuality, VouchType};
+use frame_system::{self as frame_system, ensure_signed, pallet_prelude::OriginFor};
+use log::info;
+pub use pallet::*;
+use sp_std::convert::TryInto;
+pub use weights::WeightInfo;
+// Logger target
+const LOG: &str = "encointer::vouches";
```
clangenb, line 27: Will this be printed with `-lencointer=debug`? this would be awesome

primitives/src/vouches.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#[derive(Default, Encode, Decode, PartialEq, Eq, RuntimeDebug, Clone, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "serde_derive", derive(Serialize, Deserialize))]
+pub struct Vouch<Moment> {
+	pub protected: bool,
+	pub timestamp: Moment,
+	pub vouch_type: VouchType,
+	pub quality: VouchQuality,
+}
```
clangenb, line 64: I think that some fields of these two structs are not really self-explanatory. Some documentation would be nice. :)

vouches/src/lib.rs
clangenb: I am not sure either.

vouches/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+	#[pallet::storage]
+	#[pallet::getter(fn vouches)]
+	pub(super) type Vouches<T: Config> = StorageDoubleMap<
+		_,
+		Identity,
+		T::AccountId,
+		Identity,
+		T::AccountId,
+		BoundedVec<Vouch<T::Moment>, T::MaxVouchesPerAttester>,
```
clangenb, line 102: Here I am just wondering. Is this going to be a significant number to just have a theoretical upper bound for the PoV, or do we indeed want to limit the number of vouches per attester to something like 50? Probably the former to support arbitrary use cases, but I think the latter could be interesting as well. :)

vouches/src/mock.rs
```diff
@@ -0,0 +1,51 @@
...
+
+// boilerplate
+impl_frame_system!(TestRuntime);
+impl_timestamp!(TestRuntime);
+
+// genesis values
+pub fn new_test_ext() -> sp_io::TestExternalities {
+	let t = frame_system::GenesisConfig::<TestRuntime>::default().build_storage().unwrap();
+	t.into()
+}
```
clangenb, line 51: Finally a simple pallet without many dependencies. :)

clangenb (commented):
Simple! I like it, only some minor questions and comments.

primitives/src/vouches.rs
brenzi: I wanted a way that gives us backward compatibility while still allowing to maintain meaningful order. So I left gaps to leave space for more differentiation

vouches/Cargo.toml
brenzi: no, that's copy paste, thx

vouches/src/lib.rs
brenzi: yes it will. we should apply that elsewhere

primitives/src/vouches.rs
clangenb: Hmm, I think then we should do it more in line with the scale-codec api: it is better to add:

```rust
 #[codec(index = 10)]
KnownHuman,
```

brenzi: why? then we still need `repr(u8)`
it just doubles the number of lines. what's the benefit? Doesn't it do the same?

vouches/src/lib.rs
brenzi: introducing this in this PR would be inconsistent with other pallets. We should have a separate issue if we want to change that behaviour

vouches/src/lib.rs
brenzi: Yes, this is meant to bound the PoV size.
I envision to start with a limit of 32.

vouches/src/lib.rs
brenzi: changed to blake-concat

primitives/src/vouches.rs
brenzi: okok, clippy seems to agree with you

brenzi: actually, I think hard-coding for ordering doesn't actually work if we don't know what future variants might be. It might be necessary to change existing order for it to make any sense. dropped it

clangenb (approved):
Thanks a lot for all these fixes. It looks great now!

PR #358, opened 2023-10-25
prepare to publish to crates.io

### LINKED ISSUE #357: publish all productive pallets to crates.io
including all dependencies, even if forked

prerequisite for https://github.com/polkadot-fellows/RFCs/blob/main/text/0022-adopt-encointer-runtime.md

brenzi:
closes #357 (apart from calling `cargo publish`)

**published crates to crates.io as new based on this PR**

based on versions currently used for systemchain runtimes: 
 * polish crate metadata
 * internalize TestRandomness because parity didn't publish frame-support-test to crates.io. tiny redundancy
 * skip publishing for non-production crates
 * all forked dependencies have been published already

I think this can go into master because we should have no reasons anymore not to use crates.io for runtimes, nodes, collators aso.
* we could also remove the runtime from the parachain repo as parity is doing. 
* however, the `*-rpc`crates will still live in this repo but will need to be sourced from crates.io. therefore, the entire collator repo needs to depend on crates.io (according to basti that should work. all necessary crates should be up to date there)

I intend to first publish to crates.io and then merge. This feels wrong, but the problem is that the final check that these crates are ok will be including them in a runtime build, which I can only really test when the crates are actually fetched from crates.io

challenges:
* local dependencies must define a version too. parity solves that with separate release-crates-io-* branches.


primitives/Cargo.toml
```diff
@@ -11,7 +15,7 @@ codec = { package = "parity-scale-codec", version = "3.6.1", default-features =
     "max-encoded-len",
 ] }
 crc = "2.1.0"
-geohash = { tag = "polkadot-v1.0.0", git = "https://github.com/encointer/geohash" }
+geohash = { package = "substrate-geohash", version = "0.13.0" }
```
clangenb, line 18: Interesting naming, I would not have thought of that, but I think it is a great fit.

clangenb (approved):
Nothing striking, looks good to me.

brenzi:
I'll push a few changes once publish is done. Crates.io has a rate limit for new crates, so it takes a while...

PR #368, opened 2024-02-06
fixing benchmarks if genesis is pure defaults

brenzi:
this is necessary for fellows/runtimes, because no genesis spec is available to benchmarks there:
https://github.com/polkadot-fellows/runtimes/issues/155


clangenb (approved):
LGTM

PR #371, opened 2024-02-20
Ab/polkadot v1.7.0

brenzi:
nothing out of the ordinary

release as 6.0.0 to crates.io


clangenb (approved):
Gave the Cargo.lock a more thorough look, which seems good, and the rest looks good too.

PR #372, opened 2024-02-25
Adjust MeetupSizeTarget to 15

pifragile:
This is the simulation I ran on the meetup sizes with the new target:
image

### DIFF
--- a/ceremonies/src/mock.rs
+++ b/ceremonies/src/mock.rs
@@ -64,7 +64,7 @@ impl dut::Config for TestRuntime {
 	type MeetupMinSize = MeetupMinSize;
 	type MeetupNewbieLimitDivider = MeetupNewbieLimitDivider;
 	type WeightInfo = ();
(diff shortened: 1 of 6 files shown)


ceremonies/src/mock.rs
```diff
@@ -64,7 +64,7 @@ impl dut::Config for TestRuntime {
 	type MeetupMinSize = MeetupMinSize;
 	type MeetupNewbieLimitDivider = MeetupNewbieLimitDivider;
 	type WeightInfo = ();
-	type MaxAttestations = ConstU32<10>;
+	type MaxAttestations = ConstU32<15>;
```
brenzi, line 67: as assignments with 19 participants are possible, why don't we set this to >=18, i.e. 20 to be safe?

brenzi: only a test mock, but I'm not sure if this makes tests brittle

pifragile: good catch

PR #373, opened 2024-02-27
tolerate noshows without loss of reputation

### LINKED ISSUE #365: allow reusing linked reputation
The encointer protocol defines that registering as reputable is only possible with recent unlinked reputation to disincentivize noshows. Practical experience shows that this doesn't prevent noshows but rather limits community growth. As a countermeasure, we had proactively endorsed participants to ensure seats for newbies.

A better way would be to adapt the protocol and abandon noshow punishment (we can still "punish" client side by displaying noshows along with reputation)

need to check in detail, but probably it's enough to allow linked reputation here as well:

https://github.com/encointer/pallets/blob/f6e15ecfc2a7c85905f13ec49db69073c0d3a8ae/ceremonies/src/lib.rs#L149-L155

brenzi:
closes #365 

breaking change of reputation enum. needs state migration and adjustments in wallet and rpc

I don't see how the breaking change could be avoided: On the one hand, we want to tolerate noshows (reusing Reputation::VerifiedLinked for a later cycle), on the other hand, we don't want to allow reusing reputation to register several accounts for the same cycles (reusing Reputation::VerifiedLinked for the same cycle)
The only computationally efficient (lazy) way of matching both requirements is by wrapping the cindex in VerifiedLinked variant

_remark:_ I did consider wrapping a `u8` instead of `CeremonyIndexType = u32` and just apply modulo `% 256` to reduce state bloat. As long as ReputationLifetime is < 256, that would still be unambiguous. I decided to go for lower complexity and higher state size

migration strategy:
As we can't ensure a specific CeremonyPhase for the state migration to run, we should set all `VerifiedLinked` to `VerifiedLinked(cindex)` if migration happens during _Registering_ or _Assigning_ and to `VerifiedLinked(cindex + 1)` during _Attesting_. This way, the first cycle will not benefit from noshow tolerance, but we do avoid double-use of reputation

### DIFF
--- a/ceremonies/src/lib.rs
+++ b/ceremonies/src/lib.rs
@@ -150,17 +150,18 @@ pub mod pallet {
 					Self::participant_reputation(
 						(p.community_identifier, p.ceremony_index),
 						&p.attendee_public
-					) == Reputation::VerifiedUnlinked,
+					)
+					.is_verified_and_unlinked_for_cindex(cindex),
 					Error::<T>::AttendanceUnverifiedOrAlreadyUsed
 				);
 
 				ensure!(p.verify_signature(), Error::<T>::BadProofOfAttendanceSignature);
 
-				// this reputation must now be burned so it can not be used again
+				// this reputation must now be flagged so it can not be used again in the same cycle
 				<ParticipantReputation<T>>::insert(
 					(p.community_identifier, p.ceremony_index),
 					&p.attendee_public,
-					Reputation::VerifiedLinked,
+					Reputation::VerifiedLinked(cindex),
 				);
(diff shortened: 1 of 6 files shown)


ceremonies/src/lib.rs
clangenb, line 1: We should increase the storage version of the pallet now, so that try-runtime will make us not forget about it. ;)

clangenb (commented):
Looks good to me.  The only other option I see is that we could add a storage field that would track reputation linkage, but I doubt it is worth the complexity.

However, we should not forget to update the storage version of the pallet.

PR #375, opened 2024-02-27
protect bootstrapper reputation

### LINKED ISSUE #367: If registering as bootstrapper, the proof of attendance should be ignored or be refunded
I just noticed in the app that we register with the proof of attendance even if we are a bootstrapper, but the pallet does not refund it upon unregistering. So we should either refund it if the bootstrapper unregisters, or ignore it when registering, I would vote for the latter.

### LINKED ISSUE #374: adversary bootstrapper can register two accounts per cycle
because we don't require bootstrappers to have reputation and because we issue reputation to them nevertheless, a bootstrapper with recent attendance can:
1. register a fresh account as a reputable, using its reputation as a bootatrapper
2. register the bootstrapper account without ProofOfAttendance

the severity of this vulnerability is modest

brenzi:
closes #374 
closes #367 

* bootstrappers can no longer use their ProofOfAttendance to register a fresh account as this is a vulnerability
* upon unregistering, reputation is restored for bootstrappers too now

### DIFF
--- a/ceremonies/src/tests.rs
+++ b/ceremonies/src/tests.rs
@@ -89,13 +89,18 @@ fn correct_meetup_time(cid: &CommunityIdentifier, mindex: MeetupIndexType) -> Mo
 	time as u64
 }
 
-fn get_proof(
+fn get_maybe_proof_for_self(
 	cid: CommunityIdentifier,
 	cindex: CeremonyIndexType,
 	pair: &sr25519::Pair,
 ) -> Option<TestProofOfAttendance> {
 	match EncointerCeremonies::participant_reputation((cid, cindex), account_id(pair)) {
 		Reputation::VerifiedUnlinked => Some(prove_attendance(account_id(pair), cid, cindex, pair)),
+		// the following will fail upon registration if this reputation has been previously used to
+		// register a participant in the same cycle
+		Reputation::VerifiedLinked(_) => {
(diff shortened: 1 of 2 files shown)


ceremonies/src/tests.rs
```diff
@@ -1293,13 +1298,102 @@ fn register_with_reputation_works() {
 	});
 }
 
+#[test]
+fn double_registering_by_adversary_bootstrapper_fails() {
```
brenzi, line 1302: this test reproduces the attack described in #374

ceremonies/src/tests.rs
```diff
@@ -1293,13 +1298,102 @@ fn register_with_reputation_works() {
...
+			register(account_id(&zoran), cid, Some(proof)),
+			Error::<TestRuntime>::BootstrapperReputationIsUntransferrable
+		);
+		// now Alice abuses her bootstrapper privilege to try to register herself without proof
+		assert_ok!(register(account_id(&alice), cid, None));
+	});
+}
+
+#[test]
+fn register_as_bootstrapper_with_any_kind_of_reputation_after_unregister_works() {
```
brenzi, line 1328: this test reproduces the problem with non-refunded reputation for bootstrappers in #367

clangenb (approved):
LGTM

PR #376, opened 2024-02-27
implement storage migration for ceremonies v2 Reputation type

brenzi:
remainder of #373

### DIFF
--- a/ceremonies/src/migrations.rs
+++ b/ceremonies/src/migrations.rs
@@ -21,6 +21,56 @@ mod v0 {
 
 pub mod v1 {
 	use super::*;
+	use encointer_primitives::common::{BoundedIpfsCid, PalletString};
+	use encointer_primitives::communities::CommunityRules;
+	use frame_support::{Deserialize, Serialize};
+
+	#[derive(
+		Default, Encode, Decode, Copy, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen,
+	)]
+	#[cfg_attr(feature = "serde_derive", derive(Serialize, Deserialize))]
+	pub enum Reputation {
+		// no attestations for attendance claim
+		#[default]
+		Unverified,
+		// no attestation yet but linked to reputation
+		UnverifiedReputable,
+		// verified former attendance that has not yet been linked to a new registration
+		VerifiedUnlinked,
+		// verified former attendance that has already been linked to a new registration
+		VerifiedLinked,
+	}
+
+	impl Reputation {
+		pub fn migrate_to_v2(
+			self,
+			linked_cindex: CeremonyIndexType,
+		) -> encointer_primitives::ceremonies::Reputation {
+			match self {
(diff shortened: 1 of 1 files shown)


ceremonies/src/migrations.rs
```diff
@@ -220,4 +328,34 @@ mod test {
...
+			let state = v2::MigrateToV2::<TestRuntime>::pre_upgrade().unwrap();
+			let _weight = v2::MigrateToV2::<TestRuntime>::on_runtime_upgrade();
+			v2::MigrateToV2::<TestRuntime>::post_upgrade(state).unwrap();
+
+			// Check that all values got migrated.
+
+			assert_eq!(
+				crate::ParticipantReputation::<TestRuntime>::get((cid, 1), alice),
+				Reputation::VerifiedLinked(1)
+			);
```
clangenb, line 358: Mean the code is very simple, but we might want to test that the other reputation variants are unaltered.

ceremonies/src/migrations.rs
```diff
@@ -22,6 +22,52 @@ mod v0 {
...
+					encointer_primitives::ceremonies::Reputation::UnverifiedReputable
+				},
+				Reputation::VerifiedLinked => {
+					encointer_primitives::ceremonies::Reputation::VerifiedLinked(linked_cindex)
+				},
+				Reputation::VerifiedUnlinked => {
+					encointer_primitives::ceremonies::Reputation::VerifiedUnlinked
+				},
+				Reputation::Unverified => encointer_primitives::ceremonies::Reputation::Unverified,
+			}
```
clangenb, line 56: Not really necessary to change, but still more readable and less error-prone:


```rust
			match self {
				Reputation::VerifiedLinked => {
					encointer_primitives::ceremonies::Reputation::VerifiedLinked(linked_cindex)
				},
				reputation => reputation
			}
```

clangenb (commented):
Looks good to me, some tiny comments only.

ceremonies/src/migrations.rs
brenzi: hmmm, can't we trust the compiler with that?
I get your point about test coverage, but in this case I think the compiler is restrictive enough

ceremonies/src/migrations.rs
brenzi: wasn't aware of that. I'm surprised that this is considered type-safe by rustc

brenzi: I was right, it isn't ;-)

ceremonies/src/migrations.rs
clangenb: only if you correctly implemented the match statement ;)

ceremonies/src/migrations.rs
brenzi: `expected `encointer_primitives::ceremonies::Reputation`, found `migrations::v1::Reputation``

clangenb: well, enum variants are all the same type.

PR #379, opened 2024-03-01
upgrade polkadot v1.8.0

brenzi:


brenzi:
released this state as 7.0.0 to crates.io

brenzi:
Will merge this now with highes crate version being 7.2.0

PR #397, opened 2024-06-24
[democracy] several fixes and enhancements

### LINKED ISSUE #390: [democracy] proposal that is long enough in confirming can be cancelled if the lazy update is too late
This is obvious in the code, but I am not sure if this is intended because everything was correct in terms of voting.

https://github.com/encointer/pallets/blob/67ffcfb325d4db46a28e647c7cade5affa17ae2d/democracy/src/lib.rs#L422

### LINKED ISSUE #391: [democracy] store cancellation reason for display in UIs
Frontends currently have no simple way to know why a proposal was not approved (or not enacted)

### LINKED ISSUE #392: [democracy] illegal votes can be placed even though the proposal should be voted on anymore
Here, we see that the proposal state is updated after the new tally has been placed: https://github.com/encointer/pallets/blob/67ffcfb325d4db46a28e647c7cade5affa17ae2d/democracy/src/lib.rs#L294.

Wrong Behaviour:
* If the proposal is too old, it gets cancelled, but the tally is still updated even though the vote didn't count anymore. However, the behaviour isn't wrong, except for that the turnout is one more than was counted, actually.
* If the proposal is in confirming: Actual wrong behaviour. We count the vote even though we should first check if it is still passing and then maybe approve the proposal already. So the bug is that a `Nay` can put the proposal back to `OnGoing` even though it should have been approved.

Solution:
* I suggest we should simply update the proposal state as a first step and only insert the new tally the proposal is `OnGoing` or `Confirming` and then we have to update the proposal state again.

### LINKED ISSUE #393: [democracy] move proposal enaction to the start of assigning phase
The current implementation enacts at the beginning of *registering* phase. This leads to inconsistent rewards if the proposal adjusts community issued income as some may claim early (during attesting) and others during registering (late claim). 

Options:
* at start of *registering*.
   * con: see above: potentially inconsistent CII
* at start of *assigning*.
   * pro: allows swift adding of new locations (only if onCeremonyPhaseChange is executed before ceremony pallet where assignment happens: must be well documented). In emergencies, a proposal could be submitted and approved during a single registering phase and enacted for the same cycle)
   * con: none
* at start of *attesting*.
   * con: possible inconsistency if assigning had different rules than meetup evaluation

brenzi:
closes #393 
closes #392 
closes #391 
closes #390 

* needs storage migration
* will allow very lazy update: a proposal with fulfilled confirming phase will be scheduled for enactment even if update is called after its lifetime ended - unless it has been superseded

### DIFF
--- a/communities/src/lib.rs
+++ b/communities/src/lib.rs
@@ -175,6 +175,11 @@ pub mod pallet {
 			location: Location,
 		) -> DispatchResultWithPostInfo {
 			T::TrustableForNonDestructiveAction::ensure_origin(origin)?;
+			ensure!(
+				<pallet_encointer_scheduler::Pallet<T>>::current_phase() ==
+					CeremonyPhaseType::Registering,
+				Error::<T>::RegistrationPhaseRequired
+			);
 			Self::do_add_location(cid, location)
 		}
 
@@ -191,6 +196,11 @@ pub mod pallet {
 			location: Location,
 		) -> DispatchResultWithPostInfo {
 			T::CommunityMaster::ensure_origin(origin)?;
+			ensure!(
+				<pallet_encointer_scheduler::Pallet<T>>::current_phase() ==
+					CeremonyPhaseType::Registering,
+				Error::<T>::RegistrationPhaseRequired
+			);
 			Self::do_remove_location(cid, location)
 		}
 
@@ -409,11 +419,6 @@ impl<T: Config> Pallet<T> {
 		cid: CommunityIdentifier,
 		location: Location,
 	) -> DispatchResultWithPostInfo {
-		ensure!(
-			<pallet_encointer_scheduler::Pallet<T>>::current_phase() ==
-				CeremonyPhaseType::Registering,
-			Error::<T>::RegistrationPhaseRequired
-		);
 		Self::ensure_cid_exists(&cid)?;
 		Self::validate_location(&location)?;
 		let geo_hash = GeoHash::try_from_params(location.lat, location.lon)
@@ -450,11 +455,6 @@ impl<T: Config> Pallet<T> {
 		cid: CommunityIdentifier,
 		location: Location,
 	) -> DispatchResultWithPostInfo {
-		ensure!(
-			<pallet_encointer_scheduler::Pallet<T>>::current_phase() ==
-				CeremonyPhaseType::Registering,
-			Error::<T>::RegistrationPhaseRequired
-		);
 		Self::ensure_cid_exists(&cid)?;
 
 		let geo_hash = GeoHash::try_from_params(location.lat, location.lon)
--- a/democracy/src/lib.rs
+++ b/democracy/src/lib.rs
@@ -282,13 +282,7 @@ pub mod pallet {
 			let tally = <Tallies<T>>::get(proposal_id).ok_or(Error::<T>::InexistentProposal)?;
 
 			// make sure we don't vote on proposal that can't update anymore
-			match Self::do_update_proposal_state(proposal_id) {
-				Ok(_) => Ok(()),
-				Err(error) => match error {
-					Error::<T>::ProposalCannotBeUpdated => Ok(()),
-					other_error => Err(other_error),
-				},
-			}?;
+			Self::do_update_proposal_state(proposal_id)?;
 
 			let num_votes =
 				Self::validate_and_commit_reputations(proposal_id, &sender, &reputations)?;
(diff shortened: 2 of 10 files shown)


brenzi:
@pifragile FYI. I had to move on.....

democracy/src/lib.rs
```diff
@@ -416,12 +433,14 @@ pub mod pallet {
...
+				proposal.start < last_approved_proposal_for_action.unwrap().0;
 			let proposal_too_old = now - proposal.start > T::ProposalLifetime::get();
-			if proposal_cancelled || proposal_too_old {
-				proposal.state = ProposalState::Cancelled;
+			if proposal_cancelled_by_other {
+				proposal.state =
+					ProposalState::SupersededBy { id: last_approved_proposal_for_action.unwrap().1 }
 			} else {
 				// passing
 				if Self::is_passing(proposal_id)? {
```
clangenb, line 446: Complicated logic here, but it seems correct:

Passing:
1. Passing && longer than confirmation period -> approve
2. Passing && shorter than confirmation period && proposal too old -> reject
3. Passing && shorted than confirmation period && not too old -> confirming

Not confirming:
1. Not passing && too old -> reject
2. Not passing && confirming -> ongoing. Why do we check here for confirming? is this just a safety measure to not accidentally overwrite rejected/approved with ongoing?

democracy/src/migrations.rs
```diff
@@ -0,0 +1,136 @@
...
+	#[storage_alias]
+	pub(super) type CancelledAt<T: Config> =
+		StorageMap<Pallet<T>, Blake2_128Concat, ProposalActionIdentifier, u64, OptionQuery>;
+}
+
+pub mod v1 {
+	use super::*;
+
+	#[allow(dead_code)]
+	pub struct MigrateV0toV1purging<T>(sp_std::marker::PhantomData<T>);
```
clangenb, line 21: They have now the `VersionedMigration<V1, V2>` which runs the pre/post version check internally in substrate, but it doesn't really matter, this is just an FYI

democracy/src/migrations.rs
clangenb, line 1: Seems sensible to be, as it is only on gesell, we don't care about the history of proposals, and make everybody's life easier by just purging the state.

clangenb (approved):
LGTM

communities/src/lib.rs
```diff
@@ -175,6 +175,11 @@ pub mod pallet {
 			location: Location,
 		) -> DispatchResultWithPostInfo {
 			T::TrustableForNonDestructiveAction::ensure_origin(origin)?;
+			ensure!(
+				<pallet_encointer_scheduler::Pallet<T>>::current_phase() ==
+					CeremonyPhaseType::Registering,
+				Error::<T>::RegistrationPhaseRequired
+			);
```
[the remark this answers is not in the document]
brenzi: it was there before. the problem was that we can't enact proposals at the beginning of assigning if the required phase is registering.

therefore, this check had to move. But now it is our responsibility to make sure enactments happen before assignment of cycle gatherings

democracy/src/lib.rs
brenzi: > Why do we check here for confirming? is this just a safety measure to not accidentally overwrite rejected/approved with ongoing?

No. we need to make sure a proposal falls out of confirming if the approval falls below the threshold

democracy/src/lib.rs
```diff
@@ -273,6 +281,15 @@ pub mod pallet {
...
 			let tally = <Tallies<T>>::get(proposal_id).ok_or(Error::<T>::InexistentProposal)?;
 
+			// make sure we don't vote on proposal that can't update anymore
+			match Self::do_update_proposal_state(proposal_id) {
+				Ok(_) => Ok(()),
+				Err(error) => match error {
+					Error::<T>::ProposalCannotBeUpdated => Ok(()),
+					other_error => Err(other_error),
+				},
+			}?;
```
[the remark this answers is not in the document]
brenzi: oh, actually, I just realize this is not what we want ;-)
the vote extrinsic should fail early if a proposal can't be updated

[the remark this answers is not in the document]
brenzi: I think the idea was that the vote triggers an update the same way that the extrinsic update_proposal_state would, but I think this is not necessary. we can expect someone to call the specific update call, but it is not intuitive to call "vote" if the proposal has expired

democracy/src/migrations.rs
brenzi: thx

democracy/src/migrations.rs
brenzi: yes

democracy/src/lib.rs
clangenb: Yes, I just mean that we probably want to set it to `ongoing` in any case, but we can only get to that branch of the if statement if it is in confirming anyhow, otherwise, we have a bug somewhere else in the code.

democracy/src/lib.rs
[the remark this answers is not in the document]
clangenb: I think you are right about that.

PR #398, opened 2024-06-25
polkadot v1.13.0

clangenb:
Released the crates as encointer v12 major version


brenzi (commented):
I suggest to also bump:
scale_info to 2.11.3

PR #400, opened 2024-08-20
implement proposal type petition

### LINKED ISSUE #383: [democracy] add `Petition` as ProposalAction
Petitions can be used for non-binding signalling by the electorate. The payload shall be a human readable text blob. This will allow most generic use cases, even for off-chain community matters based on social contract only

brenzi:
closes #383

### DIFF
--- a/democracy/src/lib.rs
+++ b/democracy/src/lib.rs
@@ -20,20 +20,20 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use encointer_primitives::{
-	ceremonies::ReputationCountType,
-	democracy::{Proposal, ProposalAction, ProposalIdType, ReputationVec},
-	fixed::{transcendental::sqrt, types::U64F64},
-	scheduler::{CeremonyIndexType, CeremonyPhaseType},
+    ceremonies::ReputationCountType,
+    democracy::{Proposal, ProposalAction, ProposalIdType, ReputationVec},
+    fixed::{transcendental::sqrt, types::U64F64},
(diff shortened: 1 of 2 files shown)


communities/src/lib.rs
```diff
@@ -93,7 +93,8 @@ pub mod pallet {
 		///
 		/// May only be called from `T::TrustableForNonDestructiveAction`.
 		#[pallet::call_index(0)]
-		#[pallet::weight((<T as Config>::WeightInfo::new_community(), DispatchClass::Normal, Pays::Yes))]
+		#[pallet::weight((<T as Config>::WeightInfo::new_community(), DispatchClass::Normal, Pays::Yes)
+        )]
```
clangenb, line 97: Hmm, weird fmt.

clangenb (commented):
LGTM. Quite simple. 🥳

pifragile:
LGTM 👍
One thing to keep in mind: We only allow one concurrent proposal per `ProposalActionIdentifier`. This means one `Petition` per community and one global one. So if there are multiple petitions being voted on in one community, as soon as one gets enacted, the others will be cancelled. While this implementation makes sense for things like updating the demurrage, for petitions this would be debatable.

pifragile (approved):

brenzi:
Good point. We don't want to cancel petitions. Will look into it

pifragile:
> Good point. We don't want to cancel petitions. Will look into it
> 
> edit:
> 
> * there may be other proposal types which should not supersede the same type, like treasury spends. So I introduced a helper fn to define the behavior per action type

seems legit!

PR #401, opened 2024-08-27
introduce community treasuries MVP

brenzi:
Motivation:
* each community should have an account which can receive and hold funds. Spending those funds can be done through local community democracy

Tasks
* [x] add pallet-encointer-treasuries with no dispatchables, just provides helpers for other pallets and rpc for clients
* [ ] add ProposalAction::SpendNative(maybecid, beneficiary, amount) to democracy

### DIFF
--- a/treasuries/src/lib.rs
+++ b/treasuries/src/lib.rs
@@ -1,0 +1,97 @@
+// Copyright (c) 2019 Alain Brenzikofer
+// This file is part of Encointer
+//
+// Encointer is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Encointer is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use core::marker::PhantomData;
+use encointer_primitives::{communities::CommunityIdentifier};
+use frame_support::{
+    traits::{Currency, ExistenceRequirement::KeepAlive, Get, NamedReservableCurrency},
+    PalletId,
+};
+use log::info;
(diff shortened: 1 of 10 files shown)


treasuries/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+		) -> T::AccountId {
+			let treasury_identifier =
+				[<T as Config>::PalletId::get().0.as_slice(), maybecid.encode().as_slice()]
+					.concat();
+			let treasury_id_hash: H256 = T::Hashing::hash_of(&treasury_identifier).into();
+			T::AccountId::decode(&mut treasury_id_hash.as_bytes())
+				.expect("32 bytes can always construct an AccountId32")
+		}
+
+		/// returns the account id where remaining funds of closed faucets go
```
pifragile, line 77: remove comment

treasuries/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+		sp_core::H256: From<<T as frame_system::Config>::Hash>,
+		T::AccountId: AsRef<[u8; 32]>,
+	{
+		pub fn get_community_treasury_account_unchecked(
+			maybecid: Option<CommunityIdentifier>,
+		) -> T::AccountId {
+			let treasury_identifier =
+				[<T as Config>::PalletId::get().0.as_slice(), maybecid.encode().as_slice()]
+					.concat();
+			let treasury_id_hash: H256 = T::Hashing::hash_of(&treasury_identifier).into();
```
pifragile, line 72: this code is dup from the faucet pallet, i would suggest putting this into a util function `get_account_id_from_slice`

pifragile (changes requested):
lgtm overall, only minor comments.

form a high level perspective, what is the rationale behind this? like if somebody wants to support a community, they can top up their treasury with native and then the community can decide on how they want to spend it?

treasuries/src/lib.rs
clangenb: There exists a cool function in substrate that achieves that anyhow:

```rust
		pub fn treasury_account_id() -> T::AccountId {
			T::PalletId::get().into_account_truncating()
		}

```

clangenb (approved):
LGTM

treasuries/src/lib.rs
brenzi: we currently have no cross-pallet utils library. I agree that we could abstract this away, but introducing a lib seems to be an overkill.

pifragile (approved):

brenzi:
* [x] published to crates.io

PR #404, opened 2024-09-17
fix democracy benchmarks

brenzi:
outdated logic about confirmation period in blocks instead of milliseconds needed fixing


brenzi:
published democracy pallet to crates.io as 13.3.2

PR #406, opened 2024-10-21
bump to polkadot stable-202409-1 and release encointer as v14

clangenb:
released on crates.io as v14


brenzi (approved):
LGTM

I may have preferred to skip major revisions for the upstream versions we skipped. But on the other hand, we will probably never need a release which was skipped by the fellowship

clangenb:
I agree, I wasn't thinking that far, but I also agree with that we probably never want to release something that the fellowship has skipped.

I will keep it in mind though for the next time.

PR #411, opened 2025-01-12
Ab/swapnative

### LINKED ISSUE #408: [democracy] SwapNative proposals
Similar to SpendNative, but the proposer can attach an amount of CC to the proposal which gets
1. burned if the proposal is approved
2. returned if the proposal is denied

option:
* allow proposer to top up the swap CC amount while a proposal is running (Idea is that proposers can test approval levels with more beneficial swap and offer more CC to get more approval without needing to start a new proposal

brenzi:
closes #408 

introduce proposal type SwapNativeOption


treasuries/src/lib.rs
```diff
@@ -17,47 +17,129 @@
...
+		T::AccountId,
+		SwapNativeOption<BalanceOf<T>, T::Moment>,
+		OptionQuery,
+	>;
+	#[pallet::call]
+	impl<T: Config> Pallet<T>
+	where
+		sp_core::H256: From<<T as frame_system::Config>::Hash>,
+		T::AccountId: AsRef<[u8; 32]>,
+		u64: TryFrom<BalanceOf<T>>,
```
clangenb, line 94: Those trait bounds seem a bit hacky. Additionally, I just double checked; you can simply remove the `T::AccountId: AsRef<[u8; 32]>` bound.

treasuries/src/lib.rs
```diff
@@ -17,47 +17,129 @@
...
+				..swap_option
+			};
+			<SwapNativeOptions<T>>::insert(cid, &sender, new_swap_option);
+			Self::do_spend_native(Some(cid), &sender, desired_native_amount)?;
+			Ok(().into())
+		}
+	}
 	impl<T: Config> Pallet<T>
 	where
 		sp_core::H256: From<<T as frame_system::Config>::Hash>,
```
clangenb, line 145: Hmm, I actually don't understand why we want to use the `T::Hashing` trait here. We can remove those traitbounds if we directly call the blake256 here.

Additionally, I have a suggestion: https://github.com/encointer/pallets/issues/412

treasuries/src/tests.rs
```diff
@@ -72,3 +74,222 @@ fn treasury_getter_works() {
...
+		let rate = Some(BalanceType::from_num(rate_float));
+		let cid = CommunityIdentifier::default();
+		let swap_option: SwapNativeOption<Balance, Moment> = SwapNativeOption {
+			cid,
+			native_allowance,
+			rate,
+			do_burn: burn,
+			valid_from: None,
+			valid_until: None,
+		};
```
clangenb, line 93: I don't see any tests, for the validity period, but I don't see that you check it yet either in the code. What is the plan here?

treasuries/src/lib.rs
clangenb: You can also remove the u64, with some trickery, see my PR into your PR: https://github.com/encointer/pallets/pull/413

clangenb (commented):
Some nitpickery, otherwise it looks good!

balances/src/lib.rs
```diff
@@ -272,47 +274,47 @@ impl<T: Config> Pallet<T> {
...
-		dest: T::AccountId,
+		source: &T::AccountId,
+		dest: &T::AccountId,
 		amount: BalanceType,
 	) -> Result<BalanceType, DispatchError> {
 		// Early exist if no-op.
 		if amount == 0u128 {
-			Self::deposit_event(Event::Transferred(cid, source, dest, amount));
+			Self::deposit_event(Event::Transferred(cid, source.clone(), dest.clone(), amount));
 			return Ok(amount);
```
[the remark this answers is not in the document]
brenzi: The reasoning was that we usually pass AccountId as a reference because the caller of the fn should not need to know if a clone is necessary or not. and sometimes it isn't necessary. Felt more consistent

treasuries/src/tests.rs
brenzi: I'm convinced that we will need these bounds in the next step and I want to avoid a state migration. may be premature optimization, but....

balances/src/lib.rs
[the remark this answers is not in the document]
clangenb: I don't understand where this is supposed to be consistent. Additionally, the clone is always necessary internally, so I am quite positive that the best practice is to pass an owned value in this case, as passing a borrowed value, will be only equally or less efficient than passing an owned value.

treasuries/src/tests.rs
clangenb: okok

treasuries/src/lib.rs
brenzi: using associated types is what we do everywhere else. why would we specify the hasher explicitly in this case? I don't see a problem with it, but it feels unusual

brenzi: I'll leave it as it was already done in faucet. we can refactor this anytime without difficulties

clangenb: Okok, sure. maybe too much attention to the detail here. :)

brenzi:
clippy error seems out-of-scope

PR #421, opened 2025-04-01
[treasury] Harden swap native unit tests

clangenb:
It seems that https://github.com/encointer/encointer-wallet-flutter/issues/1773#issuecomment-2765386660 is coming from the app after all, but we harden the unit tests here by:

* Aligning the test types with the actual runtime types
* Add some more test cases with rs test


brenzi (approved):
LGTM

PR #422, opened 2025-04-09
Add new Proposal Types `SwapAsset` and `IssueSwapAssetOption` abstract payment mechanism behind a trait

clangenb:


treasuries/src/payout.rs
```diff
@@ -0,0 +1,54 @@
...
+	/// Make a payment and return an identifier for later evaluation of success in some off-chain
+	/// mechanism (likely an event, but possibly not on this chain).
+	fn pay(
+		from: &Self::AccountId,
+		to: &Self::AccountId,
+		asset_kind: Self::AssetKind,
+		amount: Self::Balance,
+	) -> Result<Self::Id, Self::Error>;
+
+	fn is_asset_supported(asset_id: &Self::AssetKind) -> bool;
```
clangenb, line 38: Not sure if we need this. We can just limit the asset id enum to supported values.

PR #471, opened 2026-02-21
fix parachain runtime offline-payment benchmarking

brenzi:
The chains had different ED

### DIFF
--- a/offline-payment/src/benchmarking.rs
+++ b/offline-payment/src/benchmarking.rs
@@ -183,9 +183,9 @@ benchmarks! {
 
 		let amount: BalanceOf<T> = 10u32.into();
 
-		// Fund sender and recipient above existential deposit so both accounts exist
-		T::Currency::make_free_balance_be(&sender, 1_000_000u32.into());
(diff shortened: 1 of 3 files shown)


offline-payment/src/benchmarking.rs
```diff
@@ -183,9 +183,9 @@ benchmarks! {
 
 		let amount: BalanceOf<T> = 10u32.into();
 
-		// Fund sender and recipient above existential deposit so both accounts exist
-		T::Currency::make_free_balance_be(&sender, 1_000_000u32.into());
-		T::Currency::make_free_balance_be(&recipient, 1_000_000u32.into());
+		// Fund sender and recipient well above existential deposit (33M+ on Kusama system paras).
+		T::Currency::make_free_balance_be(&sender, 1_000_000_000u32.into());
+		T::Currency::make_free_balance_be(&recipient, 1_000_000_000u32.into());
```
clangenb, line 188: You could derive the ED from either the Currency (not sure) or pallet_balances implementation, but I guess this suffices.

PR #485, opened 2026-09-23
fix unbound iteration > pop-bound

brenzi:
fixes a possibly unbounded update of personhood rings

* affects benchmarks
* risk: if the runtime upgrade lands exactly when a batch of ring updates is unfinished, the storage model breaks. However, if we are the ones to enact the upgrade, we can time it safely. The impact is negligible as the feature isn't yet rolled out in UIs so it won't affect users


clangenb (approved):
Looks good to me, although the review was shallow, as I have not looked a the reputation rings code before.

brenzi:
published as patches:
* pallet-encointer-reputation-rings v24.0.1
* pallet-encointer-ceremonies v24.0.1

